PR #10, opened 2015-07-26
Add the ability to escape commands (or directory names) with a slash

msikma:
Currently it's not possible (to my knowledge?) to run a command that contains a space. For example:

    $ concurrent "/Users/msikma/Public projects/react-redux-dada/run.js --argument"

This will fail because everything to the left of the space is considered to be the command's full path, and everything on the right is the arguments. The intent is to call everything inside the quotation marks as a command with no arguments.

Normally, in a terminal, you'd fix this by escaping the space in the path, but Concurrently splits the command by space. This PR replaces that split with one that takes escape slashes into account. Making the following possible: 

    $ concurrent "/Users/msikma/Public\ projects/react-redux-dada/run.js --argument"

This passes all tests and handles cases with and without escaped path, with and without arguments.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -104,6 +104,8 @@ function stripCmdQuotes(cmd) {
     // Removes the quotes surrounding a command.
     if (cmd[0] === '"' || cmd[0] === '\'') {
         return cmd.substr(1, cmd.length - 2);
+    } else {
+        return cmd;
     }
 }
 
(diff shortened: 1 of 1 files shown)


kimmobrunfeldt:
Now merged and released, sorry that it took so long. https://github.com/kimmobrunfeldt/concurrently/releases/tag/1.0.0

msikma:
Very cool! I had forgotten about this :) :+1:

PR #11, opened 2015-07-27
Add next-signal option to concurrently

ashtonwar:
The next-signal option allows processes to be booted in serial, allowing us to add processes to Concurrently at runtime in a controlled way. This means we can avoid race conditions and unpredicatable outcomes in our scripts.

I recorded a video of me live-coding this feature in the context of another project:

https://www.youtube.com/watch?v=b10djzVmoaw

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -66,7 +66,7 @@ function parseArgs() {
             ' disables prettifying and colors'
         )
         .option(
-            '-n, --next-signal',
+            '-n, --next-signal [nextSignal]',
             'if string, commands are launched in sequence. Launches the next process\n' +
             'when the current process logs this value. Possible values: string or false'
         )
--- a/src/main.js
+++ b/src/main.js
@@ -28,12 +28,17 @@ var config = {
     color: true,
 
     // If true, the output will only be raw output of processes, nothing more
-    raw: false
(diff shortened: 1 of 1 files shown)


kimmobrunfeldt:
Thanks for the pull request, I especially enjoyed watching the screencast!

While I like the refactoring in this PR, I'm not sold to the idea of handling race-conditions with concurrently. This sounds like a bit hackish way to achieve reliable build-pipe. I would suggest that instead of using --next-signal option, you should make the underlying commands to be safe of race-conditions. For example in a web project: 

1. Build task which makes sure the tasks are run sequentially. 
2. Watcher task which watches for file changes and runs the build task on changes. 
3. File serving task which just serves the files for browser

The you can use concurrently to run 2. and 3. tasks. 

I can't find the other library which is very similar to concurrently, but also provides `--sequential` option. I liked the API in that library. If we are adding sequential support for concurrently, I'd go with a new option like that instead of relying to finding strings from process' stdout. What do you think?

kimmobrunfeldt:
Here's the other library I was talking about: https://github.com/mysticatea/npm-run-all

kimmobrunfeldt:
Inactive. Closing.

PR #12, opened 2015-08-18
Add option to return success or failure of the first child to exit

paulerickson:
### Scenario: Run Automated Tests and Dependencies Concurrently
Given I run a webserver and a test runner with `--kill-others` and `--success first`,
When the test runner completes
Then I expect concurrently to terminate the webserver
And return the exit code of the test runner, regardless of whether the webserver shut down gracefully

----

`first` will stop my test script from returning false positives and make me very happy, but I can't think of a case for `last` that isn't totally contrived... it just seemed like if I add `first`, then I should add `last` too.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -200,7 +200,7 @@ function handleClose(streams, children, childrenInfo) {
         });
 
         if (aliveChildren.length === 0) {
-          exit(exitCodes);
+            exit(exitCodes);
         }
     });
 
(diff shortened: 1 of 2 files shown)


kimmobrunfeldt:
Thanks for the PR, I've thought about this functionality at some point too.

Could you format the code to be consistent with existing code? 4 space indentation etc.

paulerickson:
Sure thing, @kimmobrunfeldt .  I think it was just the indentation in main.js (blame vim), but correct me if I missed any other formatting points.

kimmobrunfeldt:
Now merged and released, sorry that it took so long. https://github.com/kimmobrunfeldt/concurrently/releases/tag/1.0.0

paulerickson:
Thanks!

PR #18, opened 2015-12-15
fix(kill_all_flag): kill main process after killing childs

lapanoid:
I really don't sure about PR as I am not profound linux user. But this issue is https://github.com/kimmobrunfeldt/concurrently/issues/4 a real blocker as I believe `concurrently` used most of the time with `npm run`. So with this PR I try to solve this by killing main process after trying to kill childs. At least this kills everything so seems like valid solution.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -3,6 +3,7 @@
 var Rx = require('rx');
 var Promise = require('bluebird');
 var program = require('commander');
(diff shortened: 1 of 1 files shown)


lapanoid:
Checked again and looks like it does not solve the problem at all. Closing

PR #20, opened 2016-01-09
Fix tests call of sleep, it should take seconds.

hirtie-maxim:
The bash sleep command takes second as argument, and don't know why but exit 1, exits with 0 on Windows(didn't test on linux macine).


kimmobrunfeldt:
Why did you close this?

hirtie-maxim:
@kimmobrunfeldt, this PR didn't pass the tests on linux.

PR #21, opened 2016-01-09
Fix killing of child process on Windows

hirtie-maxim:
Since windows doesn't accept signals as unix does, the child processes should be killed in a different way. Also changed so it could be passed on windows. The 'exit 1' command returns 0 on windows.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -8,6 +8,7 @@ var program = require('commander');
 var _ = require('lodash');
 var chalk = require('chalk');
 var spawn = Promise.promisifyAll(require('cross-spawn'));
+var isWindows = /^win/.test(process.platform);
 
 var config = {
     // Kill other processes if one dies
@@ -264,7 +265,11 @@ function handleClose(streams, children, childrenInfo) {
 
             // Send SIGTERM to alive children
             _.each(aliveChildren, function(child) {
-                child.kill();
+                if(!isWindows) {
+                    child.kill();
+                } else {
+                    spawn('taskkill', ["/pid", child.pid, '/f', '/t']);
+                }
             });
         });
     }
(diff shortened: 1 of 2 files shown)


kimmobrunfeldt:
Sorry this branch has now a few conflicts, I have been too late for merging this in. Could you rebase this branch with the new master?

Spike2050:
+1

filipesilva:
@kimmobrunfeldt any chance of getting this merged in? I'm really missing the windows support for a couple of projects I'm working in. Cheers!

filipesilva:
@hirtie-maxim I tried your PR on Windows 10 and it works great. 

A question though... is enforcing `'detached': true` for the processes necessary? That way it's not possible to read through the output in logs, which is a bit problematic for automation.

Perhaps it could be made an option? I'm happy enough to do that after merge, if you don't have time. I could really use this PR in.

hirtie-maxim:
@filipesilva unfortunately I couldn't find a solution to fix that. But if you can find a wa, then great!

filipesilva:
I tried just setting `detached:false` and it worked well. What was the case where detached needed to be set to `true`?

hirtie-maxim:
Did you check it on linux? I pretty sure that there was a issue with that.

filipesilva:
I did not, no. But I thought `concurrently` was already working correctly on Linux, and that this PR only fixed the windows issues.

kimmobrunfeldt:
All these issues should be fixed in master now.

kimmobrunfeldt:
These issues should be now fixed in 3.0.0-rc1. You can test by installing it: `npm i -g concurrently@3.0.0-rc1`. Please open a new issue to concurrently repo if this is not the case. I'll release the 3.0.0 soonish.

PR #24, opened 2016-02-04
Added timestamp support

danibram:
I added timestamp support based on moment module, so you can customize the output format


kimmobrunfeldt:
I really appreciate the effort, but I ended up doing the support for this in another commit: https://github.com/kimmobrunfeldt/concurrently/commit/2c785cb9743bc643dab83cfa248573c79d5e1549 because it solved another issue too.

I'm going to close this, but you will get a support for time stamps in next release.

PR #29, opened 2016-02-27
Enable more prefix styling options

OlsonDev:
Add command line options:
- names
- name-separator
- colors

Update lodash to 4.5.1 to use _.get(object, path)
- See https://github.com/lodash/lodash/wiki/Changelog
- _.pluck is now shorthanded _.map(objects, propertyName)
- _.contains was removed in favor of _.includes

Fix typo in test/test-functional.js

PR #35, opened 2016-06-02
Added name switch into readme

JakeGinnivan:
I knew this got added but had to hunt around to find #29, it's an awesome feature and this should make it more discoverable.


kimmobrunfeldt:
Thanks for the addition!

PR #38, opened 2016-06-29
Fix for nested concurrently --kill-others

daniloster:
I created this PR to fix the problem that kill-others hangs there without closing. By the way, thanks for creating this library.


daniloster:
I am closing it to add the windows fix together.

PR #39, opened 2016-06-29
Fix for nested concurrently --kill-others for MAC and windows fix for…

daniloster:
… the other open PR

Please, evaluate it and merge as soon as possible. I need that fix and it is related to the SIGTERM issue open and the windows fix PR open but not merged. Btw, thanks for creating this library. It is awesome. ;)

PR #40, opened 2016-07-03
Add an option to kill a tree of child processes

TimMurnaghan:
This is pretty much essential for working on windows as node will very often spawn via an extra layer of cmd shell - which means that the kill only goes to that and not to the underlying processes.

This may also help with your npm problems as it may be a good idea to take out the whole tree in that case - even on unix.

Given the existence of the tree-kill module the change is straightforward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ Options:
   -h, --help                        output usage information
   -V, --version                     output the version number
   -k, --kill-others                 kill other processes if one exits or dies
+  -a, --kill-all                    kill all the descendants of the other processes if one exits or dies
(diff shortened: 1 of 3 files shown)


kimmobrunfeldt:
Looks otherwise good but this PR has now some conflicts after #39 was merged. Also, I would prefer the flag to be `killTree` instead of `killAll`

kimmobrunfeldt:
I recreated this functionality here: https://github.com/kimmobrunfeldt/concurrently/pull/58 but instead of having it a separate flag, I made it the default behvior. Thanks.

PR #42, opened 2016-07-12
Send SIGTERM signal instead SIGINT

marcioj:
On OSX my processes just get killed if I use SIGTERM instead SIGINT. It's also logged `Sending SIGTERM to other processes` so I believe this is the signal supposed to be used.


kimmobrunfeldt:
Thanks

kimmobrunfeldt:
Could you confirm that this works by installing 3.0.0-dev version: `npm i -g concurrently@3.0.0-dev` ?

marcioj:
Hi @kimmobrunfeldt. I'm without a OSX to test, but a friend that have one just confirmed to me.

PR #53, opened 2016-08-17
Added custom name example

wesbos:
Took me a second to realize when using --names you also need to specify the prefix. Should help some others


kimmobrunfeldt:
Thanks, however please add this example to here too: https://github.com/kimmobrunfeldt/concurrently/blob/e6274c5d5be32bc4a5f7b3cab6e5358545839e39/src/main.js#L126

and then I can merge it.

kimmobrunfeldt:
Ok I added an example of this by myself. Thank you anyways for the PR. 🙇

Issue #55, opened 2016-09-12
Add something like kill others, but only when any process exit with non zero status


pawojciechowski:
I want to cancel build task when either client or server build fails and I want to run them concurrently.

Example usage:

**package.json**

``` javascript
{    ...
    "scripts": {
        "build": "npm run lint && concurrently --kill-when-non-zero 'npm run build:client' 'npm run build:server'"
    }
}
```

``` javascript
npm run build
```

kimmobrunfeldt:
Good idea. I'm open for a PR.

PR #64, opened 2016-10-16
Add new kill-others-if-one-fails option (#55)

pawojciechowski:


kentcdodds:
I opened up a PR that resolves the conflicts in this PR: #93

PR #66, opened 2016-10-19
Added ability to set default config in package.json

ahmednuaman:
Concurrently is great and I use it a number of times in my `package.json`. However, 99% of the time all the config is the same, so instead of having to repeat myself for every command, I therefore created this PR so that I can keep the config in one place. 

Simply use like this:

```
...
  "concurrently": {
    "killOthers": true,
    "success": "last"
    ...
  }
...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,48 @@ Examples:
 For more details, visit https://github.com/kimmobrunfeldt/concurrently
 ```
 
+You can also specify configuration options in your `package.json`:
+
+```
+"concurrently": {
+    // Kill other processes if one dies
+    "killOthers": false,
+
+    // How much in ms we wait before killing other processes
+    "killDelay": 1000,
+
+    // Return success or failure of the "first" child to terminate, the "last" child,
+    // or succeed only if "all" children succeed
+    "success": "all",
+
+    // Prefix logging with pid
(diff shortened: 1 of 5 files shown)


ahmednuaman:
I updated the README too.

kimmobrunfeldt:
Thanks for the effort! 

Do you think package.json is the correct place for this information? Most projects are using an `.projectrc` file to store this information.

ahmednuaman:
Sure, but that's a bit overkill, eg see: https://github.com/avajs/ava

ahmednuaman:
What we thinking?

gustavohenke:
I think that the best of both worlds would be to use something like [rc](https://www.npmjs.com/package/rc) + `package.json`, so one could provide configs via:

- CLI args
- Environment variables
- local `.concurrentlyrc` file
- `.concurrentlyrc` file in your home directory
- `package.json`
etc.

gustavohenke:
Since this PR is now very outdated (the codebase was completely rewritten), it can be discarded without further considerations.

New attempts to build this configuration loading will likely use Yargs' built-in features, eg
- http://yargs.js.org/docs/#api-configkey-description-parsefn
- http://yargs.js.org/docs/#api-pkgconfkey-cwd

PR #70, opened 2016-11-17
Fixed stdout when lines are cleared

aecz:
Some ANSI code are not well processed like those that clear the entire line or move the cursor around.
Still need to support moving the cursor to far left as its conflicting with the coloredPrefix.
https://en.wikipedia.org/wiki/ANSI_escape_code

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -391,31 +391,35 @@ function logError(prefix, prefixColor, text) {
     logWithPrefix(prefix, prefixColor, text, chalk.red.bold);
 }
 
-function logWithPrefix(prefix, prefixColor, text, color) {
-    var lastChar = text[text.length - 1];
-    if (config.raw) {
-        if (lastChar !== '\n') {
-            text += '\n';
-        }
+var lastChar;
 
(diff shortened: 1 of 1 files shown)


kimmobrunfeldt:
Could you provide an example what is incorrect behaviour and how this fixes it. I might be just too tired but I didn't yet get the problem this PR solves, thanks.

aecz:
Here is some steps to reproduce the issue:
```
git clone https://github.com/callemall/material-ui-webpack-example.git
cd material-ui-webpack-example
npm i
npm i concurrently
node_modules/.bin/concurrently "npm start"
```
Comparing the output with "npm start", we can see that many lines are not cleared but added. This is due to manually adding a line break to stdout when we should instead let the process add them.

aaronjensen:
@gustavohenke thanks for merging this, this cleans up the output for us quite a bit. If you get a chance could you please release a new version? We're happy to pull from git from now, but a new version would be great. Thanks! /cc @kimmobrunfeldt

gustavohenke:
Sure! I should have done this before. v3.5.1 is on its way! 🚢

PR #73, opened 2016-12-27
Drop `bluebird`.

wtgtybhertgeghgtwtg:
- Promises are native on all versions of `node` that `concurrently` supports
- `bluebird` is only used in [two](https://github.com/kimmobrunfeldt/concurrently/blob/fe1c628efeb4760fe1bf82204142f18aa7a225c1/test/utils.js#L27) [places](https://github.com/kimmobrunfeldt/concurrently/blob/fe1c628efeb4760fe1bf82204142f18aa7a225c1/test/utils.js#L31), both of which are tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,6 @@
   },
   "homepage": "https://github.com/kimmobrunfeldt/concurrently",
   "dependencies": {
-    "bluebird": "2.9.6",
     "chalk": "0.5.1",
     "commander": "2.6.0",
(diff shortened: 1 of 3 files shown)


JuanCaicedo:
Bluebird is great, but I agree that it's probably not needed right now

kimmobrunfeldt:
Good improvement, thanks. I've been defaulting to Bluebird promises for a while and I use it as a habit sometimes for no benefits.

PR #74, opened 2016-12-28
Fix killing processes in Windows gitbash

filipesilva:
Fix https://github.com/kimmobrunfeldt/concurrently/issues/67


kimmobrunfeldt:
Thanks for the fix.

filipesilva:
It's the least I can do, I use `concurrently` a lot!

PR #77, opened 2017-01-06
Switch from moment to date-fns.

STRML:
`date-fns` is modular and about a fifth the size of the very large (and slow) `moment` library.

Usage is the same, as are the tokens ([date-fns](https://date-fns.org/docs/format) / [moment](http://momentjs.com/docs/#/displaying/)).


STRML:
Amended and rebased.

gustavohenke:
Published [v3.2.0](https://github.com/kimmobrunfeldt/concurrently/releases/tag/3.2.0)!

Issue #75, opened 2016-12-29
Group or aggregate output?


JuanCaicedo:
If you use [gnu parallel](https://www.gnu.org/software/parallel/), it has one great feature of running tasks in parallel, but grouping all of the output for a each command and printing it when that command finishes. Is there a way to do this with `concurrently`? If not, it would be a great feature 😃 

### Given these files

```bash
# script-one.sh
sleep 1
echo first
```

```bash
# script-two.sh
echo second
sleep 2
echo third
```

### Using GNU parallel
```bash
$ parallel bash ::: script-one.sh script-two.sh
first
second
third
```

### Using concurrently
```bash
$ concurrently "bash script-one.sh" "bash script-two.sh"
[1] second
[0] first
[0] bash script-one.sh exited with code 0
[1] third
[1] bash script-two.sh exited with code 0
```

JuanCaicedo:
Wondering if anyone has any thoughts on this? I'm willing to make a PR, but want to make sure it would be a feature that would get accepted 😃

AndrewCraswell:
This seems pretty fundamental to me. It's a big pain point to run tasks concurrently if the output can't be aggregated and instead you get hundreds or thousands of garbled output lines that you need to piece together to make sense of.

gustavohenke:
Happy new year! This is now out in [`v7.0.0`](https://github.com/open-cli-tools/concurrently/releases/tag/v7.0.0).

PR #79, opened 2017-01-08
WIP - Add ability to merge output

JuanCaicedo:
I started work on addressing https://github.com/kimmobrunfeldt/concurrently/issues/75. This change makes this almost possible but it has a bug that I'll detail. instructions. I've made a [repo](https://github.com/juancaicedo/concurrently-group-test) that imports my changes and allows you test them.

### Changes
- Adds a `-g` or `--group` flag to enable output aggregation 
- No breaking changes

### Working scenario
- If the commands finish in succession (first command finishes first, then second, then third). This works great. 

### Bugs
- However, if a previous command runs slower than one of the commands that follow it, the output of the latter command(s) are swallowed up.

### Could you use some guidance on 
- How to tackle the bug above
- How to add tests
- Code improvements

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -45,7 +45,10 @@ var config = {
     color: true,
 
     // If true, the output will only be raw output of processes, nothing more
-    raw: false
+    raw: false,
+
+    // If true, the output will be grouped together for each command
+    group: false
 };
 
 function main() {
@@ -121,6 +124,12 @@ function parseArgs() {
             'The option can be used to shorten long commands.\n' +
             'Works only if prefix is set to "command". Default: ' +
             config.prefixLength + '\n'
+        )
+        .option(
+            '-g, --group',
+            'group the output written to stdout and stderr by each command.\n' +
+            'Default: ' +
+            config.group + '\n'
         );
 
     program.on('--help', function() {
@@ -246,9 +255,16 @@ function run(commands) {
     });
 }
 
+function combineSourceStreams(context, sourceStreams, shouldGroupOutput) {
+    if (shouldGroupOutput) {
+        return Rx.Observable.concat.apply(this, sourceStreams);
+    }
+    return Rx.Observable.merge.apply(this, sourceStreams);
+}
+
 function handleOutput(streams, childrenInfo, source) {
     var sourceStreams = _.map(streams, source);
-    var combinedSourceStream = Rx.Observable.merge.apply(this, sourceStreams);
+    var combinedSourceStream = combineSourceStreams(this, sourceStreams, config.group);
 
     combinedSourceStream.subscribe(function(event) {
         var prefix = getPrefix(childrenInfo, event.child);


src/main.js
```diff
@@ -246,9 +255,16 @@ function run(commands) {
...
+    if (shouldGroupOutput) {
+        return Rx.Observable.concat.apply(this, sourceStreams);
+    }
+    return Rx.Observable.merge.apply(this, sourceStreams);
+}
+
 function handleOutput(streams, childrenInfo, source) {
     var sourceStreams = _.map(streams, source);
-    var combinedSourceStream = Rx.Observable.merge.apply(this, sourceStreams);
+    var combinedSourceStream = combineSourceStreams(this, sourceStreams, config.group);
```
JuanCaicedo: Though the config is available as a global, I chose to pass it as a parameter to this new function. I prefer this style, but I'm open to changing it

JuanCaicedo:
Tests are failing because of node v0.12, mentioned in https://github.com/kimmobrunfeldt/concurrently/issues/76

JuanCaicedo:
Does the bug maybe have to do with the fact that `merge` is also used on [line 280](https://github.com/kimmobrunfeldt/concurrently/pull/79/files#diff-1c90ff38a08209f9ebd4d05d1e43358eR280)? If I try changing that use concat though, all output of later commands gets swallowed

JuanCaicedo:
My current thought right now is that this has to do with RX not handling back pressure out of the box. [Here is a link with more information](http://reactivex.io/documentation/operators/backpressure.html)

src/main.js
gustavohenke: It's okay :)

gustavohenke:
Hi @JuanCaicedo! Thanks for this PR. As you may have noted, I'm getting started here as a maintainer :)

The node v0.12 build problem was fixed. If you push something new, the status will be green.

Regarding the tests for this feature... if you could emulate the example given at #75, I guess it would be enough.

And what do you mean by the output being "swallowed"? Does this means it will never show up?

JuanCaicedo:
@gustavohenke thanks for your work, excited to see this project moving forward! Yes, that's what I mean

gustavohenke:
Okay, so I might have found a way to fix that problem.

A process with index `i` needs to buffer until all processes from index 0 to `i` (inclusive) close.
This means:
- Process 0 buffers until 0 closes;
- Process 1 buffers until 0 and 1 closes;
- Process 2 buffers until 0, 1 and 2 closes;
- ...and so on.

Possible with some wise usage of `buffer` operator together with the close streams.

gustavohenke:
Of course, there may be easier/simpler ways. I'm not a heavy user of Rx...

JuanCaicedo:
@gustavohenke I have a some more time over these next few weeks to work on this, I'll take another stab at it 😃

alerkesi:
Guys, which status for this PR? How can I help something to push this faster?

JuanCaicedo:
@alerkesi Do you want to prototype an approach like @gustavohenke mentioned [above](https://github.com/kimmobrunfeldt/concurrently/pull/79#issuecomment-275574368)? Or if you can think of a different way to do it, that would be good too (I'm also not too experienced with Rx)

PR #81, opened 2017-01-17
Upgrade spawn-default-shell to 2.0.0

purusho:
I think we need to upgrade `spawn-default-shell` to `2.0.0` to be able to use `-l` flag and read from `~/.zprofile`

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "lodash": "^4.5.1",
     "moment": "^2.11.2",
     "rx": "2.3.24",
-    "spawn-default-shell": "^1.1.0",
+    "spawn-default-shell": "^2.0.0",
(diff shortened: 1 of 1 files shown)


JuanCaicedo:
Travis is passing with the exclusion of node v0.12, like a few other PRs. I mentioned this in https://github.com/kimmobrunfeldt/concurrently/issues/76. 

This PR looks good 😃

gustavohenke:
Since #83 dropped support for node v0.12, I'll accept this PR, no need to worry about if it's working in that old version!

gustavohenke:
Published [v3.2.0](https://github.com/kimmobrunfeldt/concurrently/releases/tag/3.2.0)!

PR #84, opened 2017-01-25
Fixes #33 - Add parameter/argument passthrough option

### LINKED ISSUE #33: Add parameter/argument passthrough option
Running `npm run dev -- --foo=bar` will result in `concurrently --raw "npm run start" "gulp watch" "--foo=bar"` which doesn't properly add the parameter to any of the commands.

The double dash `--` parameter passing is part of `npm run` so "you can use custom arguments when executing scripts", https://docs.npmjs.com/cli/run-script

``` json
{
  "scripts": {
    "dev": "concurrently --raw --parameter-passthrough \"npm run start\" \"gulp watch\""
  }
}
```

---

We could add an`--parameter-passthrough` option to `concurrently` that would add the last command chunk onto each command piece. The last chunk that would be considered parameters could even be separated by `--`.

So running `npm run dev -- -- --foo=bar` would result in the following after `npm`, `concurrently --raw "npm run start" "gulp watch" "--" "--foo=bar"` and then finally `npm run start --foo=bar && gulp watch --foo=bar`

An issue is adding the [extra `--` for npm run arg passing](https://docs.npmjs.com/cli/run-script). As you can see above, for it to be effective, it would actually need to be  `npm run start -- --foo=bar && gulp watch --foo=bar`. Perhaps the npm script should be updated to include the extra `--` though: `"dev": "concurrently --raw --parameter-passthrough \"npm run start --\" \"gulp watch\""`

---

If we wanted to get fancier could even add separation with `--0`, `--1` which would go to each indexed command chunk respectively, `npm run dev -- --0 --foo=bar --1 --qux=dorf` which would finally end up at `npm run start -- --foo=bar && gulp watch --qux=dorf`

thaggie:
Use -- to pass through arguments to the first command.

```
concurrently "echo foo" "echo bar" -- baz
```

```
echo foo baz
echo bar
```

Use the -a or --all option to pass through the arguments to all the commands.

```
concurrently "echo foo" "echo bar" --all -- baz
```

```
echo foo baz
echo bar baz
```

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -63,7 +63,7 @@ function main() {
 function parseArgs() {
     program
         .version(require('../package.json').version)
-        .usage('[options] <command ...>')
+        .usage('[options] <command ...> [-- <args ...>]')
         .option(
             '-k, --kill-others',
             'kill other processes if one exits or dies'
@@ -121,6 +121,12 @@ function parseArgs() {
             'The option can be used to shorten long commands.\n' +
             'Works only if prefix is set to "command". Default: ' +
             config.prefixLength + '\n'
+        ).option(
+            '-- <args ...>',
+            'pass through the <args> to the first command.\n'
+        ).option(
+            '-a, --all',
+            'pass through the <args> to all the commands.\n'
         );
 
     program.on('--help', function() {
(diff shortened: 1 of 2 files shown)


gustavohenke:
Just to be clear: I think it's better if we always pass the args to all processes.

PR #87, opened 2017-02-07
feat(color): add FORCE_COLOR environment variable to support chalk

### LINKED ISSUE #86: Colors from child scripts not being used in logged output
Thanks for this tool! I think it's great! I'm not getting colorized output when running concurrently on a script that is outputing stuff with chalk. Here's a repo that reproduces the issue: https://github.com/kentcdodds/chalk-concurrently-bug

output

It's pretty simple. `color-red.js` is:

```javascript
#!/usr/bin/env node
const chalk = require('chalk')

console.log(chalk.red(process.argv.slice(2).join(' ')))
```

Any ideas? This is sort of a blocker for [`p-s`](https://github.com/kentcdodds/p-s) dropping the `--parallel` option and recommending people just use `concurrently`. ([more info](https://github.com/kentcdodds/p-s/pull/94)).

kentcdodds:
Closes #86

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -8,6 +8,7 @@ var _ = require('lodash');
 var treeKill = require('tree-kill');
 var chalk = require('chalk');
 var defaultShell = require('spawn-default-shell');
+var supportsColor = require('supports-color');
 var IS_WINDOWS = /^win/.test(process.platform);
 
 var config = {
@@ -184,6 +185,9 @@ function run(commands) {
         var spawnOpts = config.raw ? {stdio: 'inherit'} : {};
         if (IS_WINDOWS) {
             spawnOpts.detached = false;
+        }
+        if (supportsColor) {
+          spawnOpts.env = Object.assign({FORCE_COLOR: supportsColor.level}, process.env)
         }
 
         var child;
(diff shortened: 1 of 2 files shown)


src/main.js
```diff
@@ -185,6 +186,9 @@ function run(commands) {
         if (IS_WINDOWS) {
             spawnOpts.detached = false;
         }
+        if (supportsColor) {
```
kentcdodds: I opted for this if statement rather than what we talked about in the issue because it makes the intent a little more clear.

kentcdodds:
You can use squash and merge feature and change the commit message yourself right in the github ui: https://github.com/blog/2141-squash-your-commits

kentcdodds:
Updated. I think this is correct...

boneskull:
Ha, I just ran into a corner case here, where we're launching Mocha as a subprocess and assertion it's not dumping colors into a non-tty session.  I had to explicitly unset `FORCE_COLOR`.

Is it not possible to consume the `--color` flag that `supports-color` detects?  Granted, I don't love the idea of a library looking for a flag in this manner...

PR #92, opened 2017-02-18
Fix --kill-others flag to not incorrectly kill passing commands

### LINKED ISSUE #91: exit code `null` when --kill-others specified
This one's a little odd, but occasionally (not always), concurrently will receive an exit code of `null` from one of the processes when I specify `--kill-others` (like [here](https://travis-ci.org/kentcdodds/nps/builds/202870835)).

This may be related to #37 and #80 which also indicate an exit code of `null`. Anyone know how that could happen? And better yet, how to fix it? [Here's the code](https://github.com/kimmobrunfeldt/concurrently/blob/master/src/main.js#L272) where the exit code comes from.

For now, the workaround is to not use `--kill-others`, but that's kinda annoying... :-/ I'm happy to submit a fix if someone can give me an idea of how this is even possible.

kentcdodds:
Closes #91

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -266,6 +266,7 @@ function handleClose(streams, children, childrenInfo) {
     var exitCodes = [];
     var closeStreams = _.map(streams, 'close');
     var closeStream = Rx.Observable.merge.apply(this, closeStreams);
+    var othersKilled = false
 
     // TODO: Is it possible that amount of close events !== count of spawned?
     closeStream.subscribe(function(event) {
@@ -284,21 +285,23 @@ function handleClose(streams, children, childrenInfo) {
         if (aliveChildren.length === 0) {
             exit(exitCodes);
         }
+
+        if (config.killOthers && !othersKilled) {
+            othersKilled = true
+            // Give other processes some time to stop cleanly before killing them
+            var delayedExit = closeStream.delay(config.killDelay);
+
+            delayedExit.subscribe(function() {
+                logEvent('--> ', chalk.gray.dim, 'Sending SIGTERM to other processes..');
+
+                // Send SIGTERM to alive children
+                _.each(aliveChildren, function(child) {
+                    treeKill(child.pid, 'SIGTERM');
+                });
+            });
+        }
     });
 
-    if (config.killOthers) {
-        // Give other processes some time to stop cleanly before killing them
-        var delayedExit = closeStream.delay(config.killDelay);
-
-        delayedExit.subscribe(function() {
-            logEvent('--> ', chalk.gray.dim, 'Sending SIGTERM to other processes..');
-
-            // Send SIGTERM to alive children
-            _.each(aliveChildren, function(child) {
-                treeKill(child.pid, 'SIGTERM');
-            });
-        });
-    }
 }
 
 function exit(childExitCodes) {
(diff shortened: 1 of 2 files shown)


src/main.js
```diff
@@ -284,21 +285,23 @@ function handleClose(streams, children, childrenInfo) {
         if (aliveChildren.length === 0) {
             exit(exitCodes);
         }
-    });
 
-    if (config.killOthers) {
```
kentcdodds: The problem was that this was outside of the `closeStream.subscribe` and there was no check as to whether the exit code was non-zero. This basically meant that `--kill-others` would kill all other commands after the first exits.

I've moved this to inside of `closeStream.subscribe` with a check against the `exitCode` (also had to add a check for whether `othersKilled` was set because async (don't want to sigterm other commands twice).

kentcdodds:
I've verified this fixes my use case here: https://github.com/kentcdodds/nps/pull/111

kentcdodds:
Passing build: https://travis-ci.org/kentcdodds/nps/builds/202938500

src/main.js
gustavohenke: The `--kill-others` flag states "kill other processes if one exits or dies".
So testing the exit code should probably not be done, otherwise we may cause more breakage.

kentcdodds: I see, in that case it sounds like we need a new feature that will allow us to say we should kill processes if one of them exits with a non-zero exit code. If you're interested in a future like that then I'm happy to make a port request for that. What would you like the flag to be called?

gustavohenke: If you can undo the exit code test, I'll adapt #64 and merge it

kentcdodds: Sure!

kentcdodds:
Updated. I'm super excited to get this and #64 merged!

gustavohenke:
Hi @kentcdodds, while adapting #64, I started reflecting your solution, and I found some problems.
Because we subscribe to the close stream (with a one second delay) inside of the close stream, killing other processes will only take effect after the 2nd process closes.

Here are some options for us now:
- Set up a naive `setTimeout` timer instead of using RxJS (kill delay functionality is kept)
- Remove kill delay completely, so no further headaches around this
- Figure out the real reason behind `null` exit code, and fix it considering the code before this PR.

kentcdodds:
I vote for option 2. I can't think of any use cases for delay...

> Figure out the real reason behind `null` exit code, and fix it considering the code before this PR.

The reason is because that's the exit code when a process has been closed with `SIGTERM` which will be resolved when #64 is merged I believe.

gustavohenke:
> I can't think of any use cases for delay

The comments in the code speak about exiting cleanly. The only reference about it is a comment on #7.

> The reason is because that's the exit code when a process has been closed with `SIGTERM`

From what you mentioned here, I understand that `SIGTERM` will result in processes exiting with code `null`.
Is that correct?

kentcdodds:
> From what you mentioned here, I understand that SIGTERM will result in processes exiting with code `null`.

Correct.

gustavohenke:
Ah, cool. I finally discovered that this is [documented behavior](https://nodejs.org/api/child_process.html#child_process_event_exit), but probably only on Unix machines - I own a Windows PC, and always get exit code 1...

kentcdodds:
Cool. Ok, let me know what I can do to help get #64 merged. I really want this behavior. Thanks!

Oh, and just for a use case for that... I have a pre-commit hook on many of my projects that run linting, build, tests, etc. concurrently. If any of those fail then my hook fails. I don't want to wait until they all finish before I see that one failed, so I want to kill all processes as soon as the first fails.

PR #93, opened 2017-02-27
Add new kill-others-on-fail option (#55)

kentcdodds:
This rebases and resolves merge conflicts from @pawojciechowski's PR #64

I really want to get this in. Please let me know what more needs to happen to get this merged :)

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -15,6 +15,9 @@ var config = {
     // Kill other processes if one dies
     killOthers: false,
 
+    // Kill other processes if one exits with non zero status code
+    killOthersOnFail: false,
+
     // How much in ms we wait before killing other processes
     killDelay: 1000,
 
@@ -68,6 +71,10 @@ function parseArgs() {
         .option(
             '-k, --kill-others',
             'kill other processes if one exits or dies'
+        )
+        .option(
+            '--kill-others-on-fail',
+            'kill other processes if one exits with non zero status code'
         )
         .option(
             '--no-color',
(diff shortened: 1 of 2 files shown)


src/main.js
```diff
@@ -285,23 +297,29 @@ function handleClose(streams, children, childrenInfo) {
         if (aliveChildren.length === 0) {
             exit(exitCodes);
         }
-
-        if (config.killOthers && !othersKilled) {
-            othersKilled = true
+        if (!othersKilled) {
+          othersKilled = true;
```
gustavohenke: By updating the flag here, you're preventing `--kill-others-on-fail` to work when the first process exists with success

kentcdodds: Hmmm... Could you elaborate? I don't think I follow. One of the tests demonstrates that this is not the case (unless I'm mistaken). What _should_ the code look like?

gustavohenke: 1. concurrently is run with `--kill-others-on-fail` set;
2. First process exits with success. Sets `othersKilled` to `true`, but doesn't kill others because exit code is 0;
3. Second process crashes. `--kill-others-on-fail` does nothing, because the control flag `othersKilled` is `true`.

kentcdodds: Updated!

kentcdodds:
🎉 wahoo! Thanks! Can't wait to use this!

gustavohenke:
I forgot to notify you, @kentcdodds.
v3.4.0 is published 😉

kentcdodds:
Thanks! I've already put it to [good use](https://github.com/kentcdodds/nps-utils/blob/69bc4bc0f320defccb5e1bcfe94081f1daba3565/src/index.js#L97)! Works great :D

PR #97, opened 2017-03-04
Default --prefix to "name" when --names is set

enaeseth:
I found it confusing that specifying `--names` without changing the `--prefix` continues to use a default prefix of `index`. This PR changes the default prefix to be `name` when `--names` is set and `index` when it's not.

It also cleans up the test code; since this adds new tests that use `readline`, I moved that functionality into `tests/utils.js`.

(cf: kimmobrunfeldt/concurrently#53)

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -24,7 +24,7 @@ var config = {
 
     // Prefix logging with pid
     // Possible values: 'pid', 'none', 'time', 'command', 'index', 'name'
-    prefix: 'index',
+    prefix: '',
 
     // List of custom names to be used in prefix template
     names: '',
@@ -58,6 +58,8 @@ function main() {
 
     parseArgs();
     config = mergeDefaultsWithArgs(config);
+    applyDynamicDefaults(config)
+
     run(program.args);
 }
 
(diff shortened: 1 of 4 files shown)


src/main.js
```diff
@@ -81,7 +83,7 @@ function parseArgs() {
             '-p, --prefix <prefix>',
             'prefix used in logging for each process.\n' +
             'Possible values: index, pid, time, command, name, none, or a template. Default: ' +
-            config.prefix + '. Example template: "{time}-{pid}"\n'
+            'index. Example template: "{time}-{pid}"\n'
```
gustavohenke: What if we mentioned both defaults? Too fancy?

jeremejevs:
@gustavohenke You should release this 🙂 Took me a couple of minutes to realize that the latest release is older than this PR.

RichDonnellan:
@gustavohenke I agree with @jeremejevs. I spent far too much time debugging only to realize that `master` isn't the same as the `3.4.0` release. 😕

gustavohenke:
@RichDonnellan @jeremejevs @enaeseth this was just published as v3.5.0.
I'm so sorry for taking this long to publish, folks!

PR #98, opened 2017-03-07
Add --sigkill option to use sigkill instead of sigterm

fabiosantoscode:
Also, exit the `concurrently` process when all the signals have been sent.


fabiosantoscode:
Closing this, as my issue was with docker not having a pid 0 and me leaving a zombie process behind. I'll reopen a PR for the first commit here though

Issue #48, opened 2016-07-28
Request: restart process


danfma:
Hey guys,

It would be nice if the concurrently could restart a process which died. This would allow in some cases, when you use build tasks or launch processes that could die but can be restored.
# Propose
- Add a --allow-restart flag to allow this behavior;
- Add a --restart-after XXX flag to set a delay time to respawn the process;
- Add a --restart-tries XXX flag to limit the number of respawn tries.

kimmobrunfeldt:
I agree this could be useful. I'm open to a PR for this.

gcca:
Currently someone is working on this feature. I open a PR for this.

gustavohenke:
Implemented by #100

PR #100, opened 2017-03-17
Restart process with non-zero status code (#48)

gcca:
Restart a process which died with non-zero exit status. 

- `--allow-restart` flag to allow this behavior;
- `--restart-after <miliseconds>` flag to set a delay time to respawn the process;
- `--restart-tries <times>` flag to limit the number of respawn tries.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -46,7 +46,16 @@ var config = {
     color: true,
 
     // If true, the output will only be raw output of processes, nothing more
-    raw: false
+    raw: false,
+
+    // If true, the process restart when it exited with status code non-zero
+    allowRestart: false,
+
+    // By default, restart instantly
+    restartAfter: 0,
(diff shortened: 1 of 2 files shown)


kepi0809:
would it be possible to restart exit status 0 as well? I'd run with `concurrently` multiple `kubectl` `port-forward`s and the idea would be to restart them when the connection to the pod is lost (this exits with code 0). Unfortunately, it won't be restarted I tried with `--allow-restart --restart-tries 5` and other combinations as well

gustavohenke:
Yep @kepi0809, we can think about this. It'll likely have to be a different flag.

Another CLI I know of (and use frequently) that can do similar is [node-dev](https://www.npmjs.com/package/node-dev) with the `--respawn` flag.

PR #102, opened 2017-04-06
Support for loading configuration from file with --config option

durad:
This PR adds functionality for --config option. It allows users to list all commands in a json config file and use it as a project. With many commands that need to be run concurrently, each with its own set of parameters, changing them on a command line and matching their name and color can become cumbersome task. Reuse of such setup is current;y possible only through batch files. Config file enables users to edit commands on a separate editor lines each close to its name and color. It also enables re-usability and versioning of such setup.


gustavohenke:
Since this PR is now very outdated (the codebase was completely rewritten), it can be discarded without further considerations.

New attempts to build this configuration loading will likely use Yargs' built-in features, eg
- http://yargs.js.org/docs/#api-configkey-description-parsefn
- http://yargs.js.org/docs/#api-pkgconfkey-cwd

Issue #101, opened 2017-03-31
Node API version?


JuanCaicedo:
I'm currently trying to parallelize something inside of a node script and was wondering if concurrently was a good candidate. I've tried [async.parallel](https://caolan.github.io/async/docs.html#parallel), but have been having some problems with it.

Right now to use concurrently I would have to create a child process and pass my command as a string. It would be great if there was a way to require concurrently and use it directly.

I think this would involve extracting most code from `src/main.js` into another file, like `src/lib.js`. Then `src/main.js` could call `main()` while `src/api.js` would require `src/lib.js` and export the public API methods. Also the `main` entry in [package.json](https://github.com/kimmobrunfeldt/concurrently/blob/master/package.json#L5) would need to point to `src/api.js`, since that's the entry that node's `require` looks at.

gustavohenke:
Yep. This would be a great addition 👍

borislemke:
I came across this requirement too, the scripts in `package.json` would become to long and made the file look polluted, this would be a superb addition. I need to launch multiple apps (up to 10) to work on my micro-service app locally.

Something like:
```javascript
import concurrently from 'concurrently'

const commands = [
    "nodemon script.js --someReallyLongParamtersMaybeEvenMultiple",
    "nodemon script2.js --someReallyLongParamtersMaybeEvenMultiple",
    ... and so on
]

concurrently(commands)
```

Would be perfect!
By the way, thanks a lot for making this module, been much help!

filipesilva:
@kimmobrunfeldt @gustavohenke I haven't yet contributed much to the repo, I think I'll start by implementing this. I've personally had the need for it as well.

CzBuCHi:
@filipesilva  i sah no progress on this so i did make my own pull request #112 ...

PR #112, opened 2017-06-28
node api - see issue #101

CzBuCHi:
mocha failed on my windows machine so before anything else i did these edits to make tests green:
  - windows doesnt have command 'sleep' - created one at './test/support/sleep.js' 
  - 'Skipping SIGINT/SIGTERM propagation tests ...' is writen to console only once 
---
main changes:
- extracted most of code to lib.js (main.js contains ony cli arguments parsing and calling run method from lib)
- moved 'exit' function in 'lib.js' to 'config' object - otherwise it would kill process, who uses api (api is overriding it)
- modified 'lib.js' 'handleClose' method so it returns exits codes in correct order 
- added api.js with single exported function - 'run(commands, options)' - see 'typings.d.ts' for details
- added typings.d.ts for typescript support 
---
PS: most of original of code was moved from main.js to lib.js => diff is little funky - try diff between old main.js and new lib.js if u wanna see changes in lib.js 
PSS: i didnt write any usage to documentation, because english is not my first language ...


CzBuCHi:
just found a bug :( ... starting investigation...

filipesilva:
Heya, just wanted to say that I had started on something a few months ago, but never really progressed much with it. If there's anything you can use from https://github.com/filipesilva/concurrently/tree/node-api (just the one commit), feel free to take it.

Also nice work on the typings!

PR #122, opened 2017-08-31
Support for loading configuration from file with --config option + default config file

dfadev:
fixed merge conflicts from #102 and added support for a default config file named **concurrently.json** when no arguments are specified


src/main.js
```diff
@@ -152,6 +162,10 @@ function parseArgs() {
             '--restart-tries <times>',
             'limit the number of respawn tries. Default: ' +
             config.restartTries + '\n'
+        ).option(
+            '--config <path>',
+            'Path to a json configuration file.\n' +
+            'This option can be used to load all option values from a provided json file.\n'
```
gustavohenke: We should just note that the default is `concurrently.json`

README.md
```diff
@@ -48,6 +48,8 @@ In package.json, escape quotes:
 
 Good frontend one-liner example [here](https://github.com/kimmobrunfeldt/dont-copy-paste-this-frontend-template/blob/5cd2bde719654941bdfc0a42c6f1b8e69ae79980/package.json#L9).
 
+With no parameters, concurrently will try to load it's configuration from **concurrently.json**.
```
gustavohenke: By applying the change suggested in https://github.com/kimmobrunfeldt/concurrently/pull/122/files#r144599446 we can remove this one (or at least fix it - `it's` => `its`!)

gustavohenke:
☝️ also if you have another point of view, let me know!

dfadev:
OK, seems to merge args now.  Could possibly be smarter about merging args but I think arguments specified on command line should override the config file.

Issue #121, opened 2017-08-23
Nodemon restart with 'rs' not working


MidnightP:
Great module!

One issue though it seems.

When I start my create-react-app and express server with:

```concurrently \"react-scripts start\" \"nodemon server/main.js\```

Restarting the server with nodemon by typing `rs` doesn't work anymore. I tried starting concurrently in reversed order bu no avail.

Any ideas?

concurrently: ^3.5.0
nodemon: ^1.11.0

**Comment by gustavohenke** [comment]
Unfortunately, concurrently does nothing to the processes' stdin.
That's why typing `rs` won't work - it's never sent to the nodemon process!

**Comment by SaarLevy** [comment]
I would also have loved to see this feature, but I guess the hassle this module saves is work typing in `Ctrl +c + npm start`

**Comment by MidnightP** [comment]
is worth?

**Comment by Jackclarify** [comment]
@pgraham https://github.com/kimmobrunfeldt/concurrently/pull/128, this pr works without '-r'(--row), `concurrently -k -r "ng build -w -dop=false" "gulp run-dev"`,it throw error: 
```
C:\pj\node_modules\concurrently\src\main.js:305
            target.stdin.write(line);
                        ^

TypeError: Cannot read property 'write' of null
    at Socket.process.stdin.on (C:\pj\node_modules\concurrently\src\main.js:305:25)
    at emitOne (events.js:115:13)
    at Socket.emit (events.js:210:7)
    at addChunk (_stream_readable.js:266:12)
    at readableAddChunk (_stream_readable.js:253:11)
    at Socket.Readable.push (_stream_readable.js:211:10)
    at Pipe.onread (net.js:585:20)
npm ERR! code ELIFECYCLE

```
https://github.com/kimmobrunfeldt/concurrently/issues/115, this answer solve my question.

**Comment by demisx** [comment]
I am running the latest 3.6.0 and node 10.4.1. Have this npm script that launches a few watch jobs in parallel:

```sh
$ concurrently --raw 'npm run watch:build' 'npm run watch:apidoc' 'npm run watch:test:unit' 'npm run watch:server'`
```

The watch mode starts fine for all. However typing "rs" to restart nodemon and hitting "Enter" (or simply hitting "Enter") started to crash with the same type of error and leaving npm/node zombie processes behind that I need to kill manually:

```sh
TypeError: Cannot read property 'write' of null
[...]

PR #128, opened 2018-01-02
Stdin handling

pgraham:
Handle input on stdin of the main process by directing it to one of the child processes.

Input will be sent to either a command specified in the input or to a default which is specified when concurrently starts.

### DIFF
--- a/test/utils.js
+++ b/test/utils.js
@@ -27,6 +27,8 @@ function run(cmd, opts) {
         readLines(child, opts.onOutputLine);
     }
 
+    readLines(child, (l) => { console.log(l); }, 'stderr');
+
     return new Promise(function(resolve, reject) {
         child.on('error', function(err) {
             reject(err);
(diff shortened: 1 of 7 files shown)


JaKXz:
This makes sense to me at a high level and would be great for #121 and others like it.

reohjs:
Hi, is this ready to merge?

densk1:
Would be great to see this committed. Very handy for re-running tests etc.

test/utils.js
```diff
@@ -28,9 +28,7 @@ function run(cmd, opts) {
         readLines(child, opts.onOutputLine);
     }
 
-    if (opts.onErrorLine) {
-        readLines(child, opts.onErrorLine, 'stderr');
-    }
+    readLines(child, (l) => { console.log(l); }, 'stderr');
```
gustavohenke: What was the reason for this change?

I see `onErrorLine` isn't used at all, but don't think we should get rid of it.

qraynaud: @pgraham do you have some time to look into this? I can't wait for this change to happen ^^ ;)

pgraham: I had previously added this option to log any output on stderr for specific tests but decided it would be better to always do this and so removed the options since it was now unused. But I agree, it would be useful to keep so I've re-added it.

konoufo:
So does this supposedly solve #121 ?

gustavohenke:
Yes, it does, @konoufo, thanks for heads up.
I'm looking to release this as soon as possible 🏃 ⏳ 🕙

PR #131, opened 2018-01-21
Removed brackets and spacing for --prefix none

### LINKED ISSUE #123: Disable prefix
Is there any way to disable the prefix completely?

JoshuaKGoldberg:
Now, the wrapping []s and spaces are only added if the prefix resolves to something other than an empty string.

Fixes #123


JoshuaKGoldberg:
Will do!

JoshuaKGoldberg:
I'm not seeing these test failures locally :(

JoshuaKGoldberg:
Yeah I don't have the time to push this through. Visually it's more pleasing, but not truly "necessary".

Hopefully if someone does have a real need for the feature, this PR can serve as a reference.

PR #139, opened 2018-04-13
Prefix: Handle line ending in \r as well as \n

chinchiheather:
This fixes an issue I had where some output was not being prefixed correctly due to the previous line ending in a carriage return rather than a new line


src/main.js
```diff
@@ -494,31 +494,36 @@ var lastChar;
...
 function logWithPrefix(prefix, prefixColor, text, color) {
 
-     if (config.raw) {
+    if (config.raw) {
         process.stdout.write(text);
         return;
     }
 
     text = text.replace(/\u2026/g,'...'); // Ellipsis
+    text = text.replace(/\r/g, '');
```
gustavohenke: Would it still work for you if `\n\r` was replaced directly, instead of removing all `\r`?

paescuj:
Closing this due to inactivity and also because I think it would be better to address this in the corresponding tool as the correct LF is `\r\n` (as opposed to `\n\r`).

Issue #138, opened 2018-04-10
Option to hide command output


kopach:
Would be great to have an option to hide output of some process. So, e.g. by running 2 processes I only see output of one of them. I can do something like this ` &> /dev/null` but it won't work on Windows OS.

Maybe as additional feature to this, would be an ability to set log level. e.g. show only error output.

gustavohenke:
Thanks for the suggestion, @kopach!

gustavohenke:
Duplicate of #25

akauppi:
Windows itself is going way towards WSL2, and recommending it for node development.

This might open a bag of worms, but I see the reasonable footprint to support to be the Windows + WSL2 combo. If so, this would allow simplification on the side of `concurrently` code base.

mokone91:
HI! its seems to be merged but not published yet..

gustavohenke:
Published in [`v6.4.0`](https://github.com/open-cli-tools/concurrently/releases/tag/v6.4.0)!

Issue #25, opened 2016-02-15
Silent mode


ElegantScripting:
Can you make a silent mode?

When using concurrently for build systems, its prompt gets a bit annoying.

Thanks.

kimmobrunfeldt:
How would you want the silent mode to work? Completely turn off output?

gaggle:
I'd love a "silent stdout" arg that works like name, color, etc. So I can run multiple processes but silence those I don't care about.

0xcaff:
If you don't care about cross compatibility, you can do the following with modern bash:

    (npm run testrpc >& /dev/null) & npm run test

PowerSupply:
It would be great to be able to specify the processes (by name/index) to silence stdout of

Christilut:
Same here. Running `tsc --watch` and `ava --watch` at the same time, I don't care about the typescript compilation spam. Would be great if I can select only one command to be silent. Or even better, only show error output.

Maybe such options are nice to add to #102 ?

gustavohenke:
If you are reading this, `--hide` flag is a thing since v6.4.0.
Run `concurrently --help` or see #173 to understand more.

PR #173, opened 2018-10-23
Add option and CLI flag to hide command output

firoxer:
Hi,

I noticed #138 and thought it might be doable. Here's a proof-of-concept that does the job as proposed by @gustavohenke in the aforementioned issue.

This is not a complete PR because some edge cases might be missing and the documentation is incomplete for this part. I'll get to them if I get an OK on the POC.

Issue #25 is affected by this PR as well, as the proposed `--hide` flag allows for implementing `--silent/-s` more easily.

Thanks for creating concurrently!


firoxer:
>What else do you think should be covered by your PR?

I was just pointing out that I'm not confident that all use cases have been covered. :) If you feel it's complete then it's all good.

I added some functional tests and they seem to work fine. While doing that, I noticed that because `--hide` uses the yargs `array` type, it makes it necessary to use `--` after the final hideable index/name when `--hide` is the last option because otherwise yargs won't know where stop parsing indices/names and where to start parsing commands. In other words, if the first and the third process of `bin/concurrently.js 'echo 1' 'echo 2' 'echo 3'` should be hidden, the command should look like `bin/concurrently.js --hide 0 2 -- 'echo 1' 'echo 2' 'echo 3'`. I expanded the usage notes to include this.

However, I was left wondering if this kind of behaviour is fine, because it kind of conflits with `--names` which works with comma-separated values instead of spaces (e.g. `--names foo,bar`, no following `--` needed). The problem is in that by telling yargs to use `array`, we get automatic casting to numbers (i.e. `--hide 1 foo 2 bar` results in an array of `[1, 'foo', 2, 'bar']`) which works really well with the rest of the code. By making `--hide` work like `--names` we lose this nicety, but we gain usage consistence. Which way do you prefer?

AntonNiklasson:
Awesome job, I would love to see this merged. However, I would expect the name of the flag to be "quiet" and not "hide". Thoughts?

firoxer:
Sure. I just went with "hide" since it was suggested in #138. I guess CLI tools usually use `-q` or `--quiet` with zero arguments so `--quiet [number]` would deviate from that, but on the other hand I don't think it's that big of a deal. If you want me to change it, I'll do it.

StefanFrederiksen:
This use-case is very ideal for me, so is there anything that still needs to be done for this PR?

firoxer:
From my part I only need to know if we want to use `--hide` or `--quiet` and I'll change the code if necessary.

StefanFrederiksen:
I think `--silent` is more verbose as to the intent of the flag. Otherwise `--quiet` since a hide flag that makes the child process not output anything doesn't seem intuitive (at least to me).

munrocket:
Very useful PR why you not merge it @kimmobrunfeldt ?

cben:
@firoxer can you resolve merge conflicts?

firoxer:
@cben Sure, done.

munrocket:
@kimmobrunfeldt @gustavohenke

shilomagen:
This is an amazing feature we're waiting for :)
Do you plan to merge it soon? Thank you so much!

gustavohenke (approved):
Wow. Nearly 3 years later... I pushed some small changes, most notably not using the yargs array type.

LGTM now, let's ship it.

gustavohenke:
Published in v6.4.0!

PR #189, opened 2019-06-05
Use prefixLength

BrunoBernardino:
The setting was apparently not being sent through via CLI.

Also added a test confirm the behavior and prevent regressions.

Please let me know if there's anything else you'd like me to do here!


gustavohenke:
Published as v4.1.1!

BrunoBernardino:
Thanks @gustavohenke !

PR #190, opened 2019-07-03
fix prefixColors

ddebasilio:
To fix an error:
` TypeError: args.prefixColors.split is not a function`


ddebasilio:
> Thanks for the PR (and sorry for my delay!!!) -- it LGTM, but I'm curious, when does the error happen?



If you define empty prefixColors or if it is not properly quoted. This makes sure that it works for when the flag is defined. Either way it is good practice to convert it toString

gustavohenke:
What's the command to replicate this that you have executed?
It may very well be a bug with `yargs`.

PR #195, opened 2019-08-22
Fix #193

### LINKED ISSUE #193: Concurrently doesn't terminate child processes on SIGHUP signal
When you start concurrently in a terminal, and then kill this terminal instance, child processes aren't get terminated.

From my understanding the reason for it - when terminal is killed process inside this terminal recieves SIGHUP instead of SIGINT or SIGTERM.

Would be good if concurrently can handle SIGHUP too and pass this signal to child processes.

leo-buneev:
Adds 'SIGHUP' to list of signals that trigger killing of child processes.

### DIFF
--- a/src/flow-control/kill-on-signal.js
+++ b/src/flow-control/kill-on-signal.js
@@ -8,7 +8,7 @@ module.exports = class KillOnSignal {
 
     handle(commands) {
         let caughtSignal;
-        ['SIGINT', 'SIGTERM'].forEach(signal => {
+        ['SIGINT', 'SIGTERM', 'SIGHUP'].forEach(signal => {
             this.process.on(signal, () => {
                 caughtSignal = signal;
                 commands.forEach(command => command.kill(signal));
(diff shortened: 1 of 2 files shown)


leo-buneev:
During testing I've discovered that my particular issue was not caused by this.
However, I still think that adding SIGHUP to list of signlas is useful - up to you if you want to merge it.

gustavohenke:
Landed on [`v5.0.0`](https://github.com/kimmobrunfeldt/concurrently/releases/tag/v5.0.0)!

cben:
I added logging to tree-kill and it seems it doesn't know what's SIGHUP?
```
[2] yarn insights-proxy exited with code SIGINT
--> Sending SIGTERM to other processes..
[1] mockdata/mockserver.py exited with code 1
--> Sending SIGTERM to other processes..
tree-kill: sending SIGINT to process 19857
tree-kill: sending SIGINT to process 19856
tree-kill: sending SIGINT to process 19855
tree-kill: sending undefined to process 19856
~/u/uhc-portal [130] $ tree-kill: sending undefined to process 19855
tree-kill: sending undefined to process 19855
```

PR #212, opened 2020-01-06
fix: bugfix for wildcard naming

### LINKED ISSUE #148: Run template commands
Hello.

I think there are some mistakes in the DOCS, because this stuff has never worked for me.

```
concurrently "npm:watch-*"

# Equivalent to:
concurrently -n js,css,node "npm run watch-js" "npm run watch-css" "npm run watch-node"

# Any name provided for the wildcard command will be used as a prefix to the wildcard
# part of the script name:
concurrently -n w: npm:watch-*

# Equivalent to:
concurrently -n w:js,w:css,w:node "npm run watch-js" "npm run watch-css" "npm run watch-node"
```

Thx.

### LINKED ISSUE #211: Wildcard naming is verbose and doesn't match docs
Same issue as #148 though I will include a better description & repro case.

**Problem**: if you use the `npm:` or `yarn:` wildcard syntax, the docs say we should get only the shortened wildcard portion of the commands as labels, but in reality this is broken and we get the full, verbose version of the commands, including the non-wildcard prefix.

Here is a bare-bones [repro example](https://github.com/transitive-bullshit/concurrently-bug-repro). It just uses the wildcard syntax to call `echo.js` which just logs its argv once a second.

Here is the current output:

```
$ concurrently yarn:start-*
$ node echo.js foo
$ node echo.js bar
[start-foo] foo
[start-bar] bar
[start-bar] bar
[start-foo] foo
[start-bar] bar
[start-foo] foo
```

Here is the expected and desired output:

```
$ concurrently yarn:start-*
$ node echo.js foo
$ node echo.js bar
[foo] foo
[bar] bar
[bar] bar
[foo] foo
[bar] bar
[foo] foo
```

This clearly does the opposite of what was intended and described in the project's readme.

#165 looks like the correct fix, but it is out of date. @gustavohenke if you'll accept an updated PR, I'm happy to resolve this for you.

Thanks!

transitive-bullshit:
Fixes #211 
Fixes #148 

See #211 for more info and a repro case.


transitive-bullshit:
@gustavohenke the coveralls issue seems to be a false positive.

gustavohenke:
Before merging, I thought of checking how v3 would behave.
It seems that
1. Empty wildcard match means no command name
2. Passing `--name` flag would prefix the wildcard match

```
$ ../src/main.js npm:foo-*-baz
[bar]
[bar] > @ foo-bar-baz /some/dir
[bar] > echo foo bar baz
[bar]
[1]
[1] > @ foo--baz /some/dir
[1] > echo foo--baz
[1]
[bar] foo bar baz
[1] foo--baz
[bar] npm run foo-bar-baz exited with code 0
[1] npm run foo--baz exited with code 0

$ ../src/main.js -n test npm:foo-*-baz
[test]
[test] > @ foo--baz /some/dir
[test] > echo foo--baz
[test]
[testbar]
[testbar] > @ foo-bar-baz /some/dir
[testbar] > echo foo bar baz
[testbar]
[testbar] foo bar baz
[test] foo--baz
[test] npm run foo--baz exited with code 0
[testbar] npm run foo-bar-baz exited with code 0
```

WDYT?

poteirard:
Any progress on this?

brokenmass:
@gustavohenke : can this be merged, please ?

sunknudsen:
@gustavohenke Thanks so much for `concurrently`. What is holding back this PR?

Looking forward to the behavior illustrated in the docs which will save precious space in the terminal.

```shell
concurrently "npm:watch-*"

# Equivalent to:
concurrently -n js,css,node "npm run watch-js" "npm run watch-css" "npm run watch-node"
```

LucasForster:
>     1. Empty wildcard match means no command name
> 
>     2. Passing `--name` arg would prefix the wildcard match
> 
>     WDYT?

Seems like the expected behavior. Would love to see this merged, I was wondering why my console output wouldn't match what is described in the docs.

And thanks for the package, of course!

gustavohenke:
Builds all failed because of Coverage, but will merge anyway since everything else is passing and I've applied the suggested changes above!

gustavohenke:
Published as v6.5.1!

PR #213, opened 2020-01-15
Return exitCodes

testerez:
Return the exit codes on success or failures so it can be easily used by the caller.
My use case is to display a clear message showing which command failed at the end of the execution.


gustavohenke:
Published as v5.1.0.

PR #228, opened 2020-06-07
Include command info and index in completion listener Promise value

### LINKED ISSUE #181: Feature Request: pass comman object and exit status to fail handler
when using concurrently using [javascript api](https://github.com/kimmobrunfeldt/concurrently#concurrentlycommands-options) in error handler we dont get any informaytion regarding failing process
pass relevant information in error object

peruukki:
Return command information and index in addition to the exit code in the command close stream and the completion listener Promise value. This way we can associate the exit codes with the commands when using `concurrently` programmatically.

Resolves #181.

I also tried if I could use another stream (something like `closeWithCommandInfo`) for just the additional information, so that I wouldn't need to change so many places, but I had some confusing errors in the binary unit tests, and in the end, having a single close stream seems more reasonable.

### DIFF
--- a/src/command.js
+++ b/src/command.js
@@ -5,11 +5,12 @@ module.exports = class Command {
         return !!this.process;
     }
 
-    constructor({ index, name, command, prefixColor, killProcess, spawn, spawnOpts }) {
+    constructor({ index, name, command, prefixColor, env, killProcess, spawn, spawnOpts }) {
         this.index = index;
         this.name = name;
         this.command = command;
         this.prefixColor = prefixColor;
+        this.env = env;
         this.killProcess = killProcess;
         this.spawn = spawn;
         this.spawnOpts = spawnOpts;
@@ -31,7 +32,14 @@ module.exports = class Command {
         });
         Rx.fromEvent(child, 'close').subscribe(([exitCode, signal]) => {
             this.process = undefined;
-            this.close.next(exitCode === null ? signal : exitCode);
+            this.close.next({
+                command: this.command,
+                name: this.name,
+                prefixColor: this.prefixColor,
+                env: this.env,
+                index: this.index,
+                exitCode: exitCode === null ? signal : exitCode,
+            });
         });
         child.stdout && pipeTo(Rx.fromEvent(child.stdout, 'data'), this.stdout);
         child.stderr && pipeTo(Rx.fromEvent(child.stderr, 'data'), this.stderr);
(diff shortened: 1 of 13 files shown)


src/command.spec.js
```diff
@@ -73,14 +73,38 @@ describe('#start()', () => {
...
+            index: 1,
+            spawn: () => process,
+        });
+
+        command.close.subscribe(data => {
+            expect(data.command).toBe('cmd');
+            expect(data.name).toBe('name');
+            expect(data.prefixColor).toBe('green');
+            expect(data.env).toEqual({ VAR: 'yes' });
+            expect(data.index).toBe(1);
```
gustavohenke, line 100: Can simplify to this with the above suggestion

```suggestion
            expect(data.command).toBe(command);
```

src/flow-control/kill-on-signal.js
```diff
@@ -16,8 +16,9 @@ module.exports = class KillOnSignal {
         });
 
         return commands.map(command => {
-            const closeStream = command.close.pipe(map(value => {
-                return caughtSignal === 'SIGINT' ? 0 : value;
+            const closeStream = command.close.pipe(map(exitInfo => {
+                const exitCode = caughtSignal === 'SIGINT' ? 0 : exitInfo.exitCode;
+                return Object.assign({}, exitInfo, { exitCode });
```
gustavohenke, line 21: Would need to change here too

src/command.js
```diff
@@ -31,7 +32,14 @@ module.exports = class Command {
...
         Rx.fromEvent(child, 'close').subscribe(([exitCode, signal]) => {
             this.process = undefined;
-            this.close.next(exitCode === null ? signal : exitCode);
+            this.close.next({
+                command: this.command,
+                name: this.name,
+                prefixColor: this.prefixColor,
+                env: this.env,
+                index: this.index,
+                exitCode: exitCode === null ? signal : exitCode,
```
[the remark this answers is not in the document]
peruukki: Sounds good, I was considering that too. 👍 I even tried keeping these command info fields in a `command` field in `Command` so any new fields would automatically be included, but that seemed to require even more changes elsewhere. 😄

How do you feel about `index`, should it be inside `command` or on its own at the top level?

src/flow-control/kill-on-signal.js
peruukki: This only overwrites the `exitCode` field, so it shouldn't need changing.

src/command.js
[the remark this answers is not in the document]
peruukki: I kept `index` at the top level as it's not in the original command fields. I pushed the changes necessary to change the shape of the returned value.

I tried again to change the code internally so that it would automatically pick up any changes in the "command info", but it seems to require changes in so many places that it's maybe not worth it.

gustavohenke:
Release as v5.3.0!

PR #231, opened 2020-06-15
logger: do not trim prefix

### LINKED ISSUE #227: Allow spaces at prefix start/end
Custom prefixes currently strip any space, tab, newline, or other white-space characters at the beginning and end of their supplied content.

This seems to be a bug, as I would expect the supplied value to persist unmodified.

Example input:
```shell
concurrently -p '\u0020{name}\u0020' -n foo,bar 'echo "foo"' 'echo "bar"'
```
Current output:
```shell
foo foo
bar bar
```
Desired output:
```shell
 foo  foo
 bar  bar
```

McAllaster:
The logger should preserve any leading/trailing white space provided in the prefix

Closes #227

Side note: I'm keeping boilerplate code here just because I see it within the rest of the logger specification. Normally I'd DRY it up a bit with test generators or a `beforeEach`, but I'm unsure if that is a welcome change.

### DIFF
--- a/src/logger.js
+++ b/src/logger.js
@@ -53,7 +53,7 @@ module.exports = class Logger {
         return _.reduce(prefixes, (prev, val, key) => {
             const keyRegex = new RegExp(_.escapeRegExp(`{${key}}`), 'g');
             return prev.replace(keyRegex, val);
-        }, prefix).trim();
+        }, prefix);
     }
 
     colorText(command, text) {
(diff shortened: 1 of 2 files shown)


gustavohenke:
Proceeded with the merge in case you don't have time/don't feel like tidying up the tests.

gustavohenke:
Release as v5.3.0!

PR #233, opened 2020-06-17
Test cleanup

McAllaster:
I had mentioned in #231 about cleanup of some tests, and this is the pull request for that cleanup.

Originally, I had intended to reduce the `logCommandText` section to use generated tests via Jest's `test.each` functionality. However, due to the way the descriptions and options are used, leaving the tests to be written as-is ends up being simpler to understand and modify.

What's left is just a cleanup of using references to the Logger prototype for descriptions instead of strings, using template literals instead of string addition, and instantiate a `Logger` instance the same way across the specification (using the `createLogger` function defined in the specification).

### DIFF
--- a/src/logger.spec.js
+++ b/src/logger.spec.js
@@ -14,7 +14,7 @@ const createLogger = options => {
     return logger;
 };
 
-describe('#log()', () => {
+describe(Logger.prototype.log, () => {
     it('writes prefix + text to the output stream', () => {
         const logger = new Logger({ outputStream });
         logger.log('foo', 'bar');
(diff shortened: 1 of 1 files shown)


src/logger.spec.js
```diff
@@ -3,20 +3,22 @@ const chalk = require('chalk');
...
     return logger;
 };
 
-describe('#log()', () => {
+beforeEach(() => {
+    outputStream = createMockInstance(Writable);
+    logger = createLogger();
+});
+
+describe(Logger.prototype.log, () => {
```
gustavohenke, line 20: How do you think that this reads better?

I have been using `#` to indicate prototype instead of `.` for non-prototype properties for a while now. Don't remember where I got this from, maybe Java 😛 

The rest of the codebase is also using this standard, so it would deviate a bit, but if I'm convinced I can replace the other occurrences 👍

McAllaster: In my experience, this description pattern helps validate that the method you are describing exists, and is updated appropriately when you make changes. An editor would validate that `Logger.prototype.log` exists, while it would _not_ validate a string being used in a description.

PR #244, opened 2020-09-15
feat: upgrade yargs (requires node 8)

FauxFaux:
Upgrade `yargs` to a version everyone else is using. This involves dropping support for node 6. Once we've done that, we can use less old jest and eslint, too!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ node_js:
   - 12
   - 10
   - 8
-  - 6
 script:
   - npm test
   - npm run lint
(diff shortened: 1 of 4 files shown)


gustavohenke (commented):
Can you also update `appveyor.yml` to not run on Node 6?

FauxFaux:
That's a bit more green, isn't it.

gustavohenke (approved):

gustavohenke:
Out in v6.0.0, thanks!

PR #246, opened 2020-10-28
Fixes(216): Add option cwd to concurrently and to command options

### LINKED ISSUE #216: Support `cwd` for commands
Right now I need to run a command in a specific directory like this (when using concurrently programatically.)

`cd path/to/directory &&  run-command.sh`

This is fine, but becomes problematic when my directory contains spaces. It would be great if we had support for specifying a current working directory.

thdk:
Fixes #216

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,11 +10,13 @@ Like `npm run watch-js & npm run watch-less` but better.
 ![](docs/demo.gif)
 
 **Table of contents**
-- [Why](#why)
-- [Install](#install)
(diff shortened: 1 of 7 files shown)


phawxby:
@gustavohenke & @kimmobrunfeldt it would be great to get this merged in and released.

gustavohenke (approved):
This is a fantastic addition, thank you!
What a shame that I took this long to look at the PRs open here...

Issue #247, opened 2020-11-04
Add support for a grid layout


lneves12:
It would be nice to have the output of each process being shown inside a grid layout.

I have been playing around with concurrently and implemented something very simple: 

https://github.com/lneves12/concurrently/commit/054cbd3dccecc42fedf8a107ad3f1dc7bc6220c6

Demo:

![video_concurrently](https://user-images.githubusercontent.com/1140720/98118561-5042f500-1ea3-11eb-9d5e-9c0d6a81e068.gif)

Just wondering if this is something you would like to have in this library. If yes, I can clean up that code and add it under a: "--grid" option.

What do you think? Any suggestion?

milahu:
node-ncurses? holy cow ^^ but maybe a browser-based interface would be prettier?

\- goldenlayout.js as window manager (also used on https://godbolt.org/ for example)
\- scroll back each process
\- select text per process

gustavohenke:
What would the output look like when there are many commands + a small terminal window?

fabiosangregorio:
I was looking for something similar to this, maybe with a tabbed layout instead of a grid, since reading the output of many processes becomes cumbersome very fast. @gustavohenke @paescuj any plan for this? 👀 

I tried [mprocs](https://github.com/pvolok/dekit/tree/master) and looks like what I'd want

gustavohenke:
I think it could be added, but it'd be a pretty large change, I think?

Depending on what's required to build the UI, maybe it'd fit best as a companion program? I remember looking at using [Ink](https://github.com/vadimdemedes/ink) for something, and it was large. Maybe not something the majority of concurrently users would have a use for, but I could be wrong - this issue itself has gotten 10 👍 so far

_If anyone's interested at taking a stab at this: do break it up into several PRs please_

PR #248, opened 2020-11-05
Adds an experimental simple grid layout

lneves12:
Issue: #247

Adds an experimental simple grid layout under an option "--grid". This assumes a grid of 2 columns, should be enough for now.

Let me know what you think about this approach. 

Example with even number of cells

Screenshot from 2020-11-05 01-47-18

Example with odd number of cells

Screenshot from 2020-11-05 01-47-03

### DIFF
--- a/bin/concurrently.js
+++ b/bin/concurrently.js
@@ -49,6 +49,12 @@ const args = yargs
                 'and concurrently coloring.',
             type: 'boolean'
         },
+        'grid': {
+            describe:
+                'Output of processes will be formatted inside a grid.\n' +
+                'Experimental layout.',
+            type: 'boolean'
+        },
         // This one is provided for free. Chalk reads this itself and removes colours.
         // https://www.npmjs.com/package/chalk#chalksupportscolor
         'no-color': {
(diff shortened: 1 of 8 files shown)


amertak:
I would love to see this feature released @lneves12

nicholas-ochoa:
Any chance of this getting merged? This would be a very useful feature when there are multiple applications running that are in a `watch` mode and outputting multiple lines at once.

paescuj:
Hey @lneves12, I'm sorry that you have not received any feedback for such a long time!
Thank you very much for your contribution! ❤️

The screenshots you've provided look awesome 💯

Nevertheless, I'm afraid we cannot include this in its current state for various reasons (besides the obvious required updates):
- In general, it seems to be to much in an "experimental state"
- [neo-blessed](https://github.com/embarklabs/neo-blessed) seems to be unmaintained - I'd like to avoid such dependencies
  (We'd need to look for an alternative, perhaps [terminal-kit](https://github.com/cronvel/terminal-kit)?)
- More tests needs to be added
- ...

That being said, this is not a general "no" to such a feature and I'd be pleased if you or someone else would like to continue working on this!

paescuj:
Closing for now for the above reasons and because the pull request is outdated. Thanks anyway! ❤️ A new approach is welcome.

PR #250, opened 2020-11-23
Include `killed` boolean in command result

Roaders:
The exit code of each command is already included in the result but the usefulness of this is limited if concurrently is run with `killOthers`. When parsing the result you do not know if a non-0 exit code is because the command failed or if it was killed by concurrently.

Adding this option allows you to print a message at the end identifying exactly which commands caused the initial failure.

### DIFF
--- a/src/command.spec.js
+++ b/src/command.spec.js
@@ -60,6 +60,7 @@ describe('#start()', () => {
 
         command.close.subscribe(data => {
             expect(data.exitCode).toBe(0);
+            expect(data.killed).toBe(false);
             expect(command.process).toBeUndefined();
             done();
         });
@@ -74,6 +75,7 @@ describe('#start()', () => {
 
         command.close.subscribe(data => {
             expect(data.exitCode).toBe('SIGKILL');
+            expect(data.killed).toBe(false);
             done();
         });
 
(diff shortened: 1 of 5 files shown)


Roaders:
Any chance we can get this PR considered?

gustavohenke (changes requested):
Hey! Can we also get some test for this, please?

Roaders:
I was going to write tests but tests were failing before I started so it's hard to know where to start.

Roaders:
As far as I can tell there are no existing tests that test the exit condition of multiple commands in the `close` stream. I am trying to find one to base my tests on.

Roaders:
OK, I figured it out. Been so long since I did this I had forgotten the structure of the project.

src/command.spec.js
```diff
@@ -164,4 +167,17 @@ describe('#kill()', () => {
 
         expect(killProcess).not.toHaveBeenCalled();
     });
+
+    it('should mark the command as killed', done => {
```
gustavohenke, line 171:
```suggestion
    it('marks the command as killed', done => {
```

gustavohenke (approved):
Thanks for the side fixes!

Roaders:
Do I have to do anything to merge this? I assume you'll merge it at some point.
Thanks.

Roaders:
Any idea when this will be released?

gustavohenke:
Heya. Just released it in v6.2.0.

Roaders:
brilliant, thanks

PR #253, opened 2020-12-25
Stop reading from stdin after programmatic API finishes

### LINKED ISSUE #252: Programmatic usage errors on Ctrl+C when specifying process.stdin
```
$ yarn node --version
v14.10.1
```
Using yarn versions 1.22.10 and 2.4.0

## Repro
1. `yarn add concurrently`
1. Add script:
    ```js
    const concurrently = require('concurrently')

    concurrently([
      'echo hi && sleep 100',
    ], {
      inputStream: process.stdin,
    }).catch(console.error)
    ```
1. `yarn node start.js` 

### With yarn 1
Doing Ctrl+C on the `yarn node start.js` command will exit, but then hitting Enter again will show an error when NodeJS tries to read the line after exiting:
```
events.js:291
      throw er; // Unhandled 'error' event
      ^

Error: read EIO
    at TTY.onStreamRead (internal/stream_base_commons.js:209:20)
Emitted 'error' event on ReadStream instance at:
    at emitErrorNT (internal/streams/destroy.js:106:8)
    at emitErrorCloseNT (internal/streams/destroy.js:74:3)
    at processTicksAndRejections (internal/process/task_queues.js:80:21) {
  errno: -5,
  code: 'EIO',
  syscall: 'read'
}
```

### With yarn 2
Run `yarn set version berry` before running `yarn node start.js` again. This time, when doing Ctrl+C, the script will hang until you do Ctrl+D

### With node
Running just `node start.js` has the same behavior as yarn 2

## Resolution

There's probably some code in `concurrently` that needs to stop listening to `inputStream` when the process exits. That should solve the problems with both yarn 1 and 2

## Workaround

Manually destroying `process.stdin` after the process exits seems to fix

```js
concurrently(...)
  .catch(console.error)
  .finally(() => {
    process.stdin.destroy()
  })
```

brandonchinn178:
Fixes #252 by explicitly calling `.pause()` on the input stream after all commands have finished. This seems to be the only way to fix the problem, and there doesn't seem to be a way to reset the `process.stdin` state completely to the state before `concurrently` ran (see [NodeJS docs](https://nodejs.org/api/stream.html#stream_three_states)).

Refactored some parts of the code in order to make this change more drop-in:
* Added `BaseHandler` class that all handlers extend from
* Add `onFinish` to `BaseHandler` which runs after all commands have finished
* Add `FakeHandler` to tests that extend `BaseHandler` instead of ad-hoc handler objects in `concurrently.spec.js`
* Use an actual `Readable` in `input-handler` tests instead of `EventEmitter`

The CLI has not changed in this PR, but the programmatic API now has two new options:
* `handleInput`: an alias for the old `inputStream: process.stdin`. Just thought this would be nice, since 99% of the time, `inputStream` is set to `process.stdin`
* `pauseInputStreamOnFinish`: Some people might have a use-case where they want to read from `process.stdin` further after running `concurrently` programmatically. In this rare instance, they'd have to explicitly set this option to `false`, or explicitly call `process.stdin.resume()` afterwards. I can't imagine this being a common use-case, so I think this explicitness is a good trade-off.

## Repro steps

1. Write `start.js` as specified in #252, except replace `require('concurrently')` with `require('.')`. Put this file in the root of the repo.
2. `node start.js` + Ctrl-C should no longer hang / wait for Ctrl-D

### DIFF
--- a/src/flow-control/base-handler.js
+++ b/src/flow-control/base-handler.js
@@ -1,2 +1,11 @@
 module.exports = class BaseHandler {
+    constructor(options = {}) {
+        const { logger } = options;
+
+        this.logger = logger;
+    }
+
+    handle(commands) {
+        return commands;
+    }
 };
--- a/src/flow-control/base-handler.js
+++ b/src/flow-control/base-handler.js
@@ -6,7 +6,7 @@ module.exports = class BaseHandler {
     }
 
     handle(commands) {
-        return commands;
+        return { commands };
     }
 
     /**
--- a/src/flow-control/base-handler.js
+++ b/src/flow-control/base-handler.js
@@ -6,11 +6,11 @@ module.exports = class BaseHandler {
     }
 
     handle(commands) {
-        return { commands };
+        return {
+            commands,
+            // an optional callback to call when all commands have finished
+            // (either successful or not)
+            onFinish: null,
+        };
     }
-
-    /**
-     * A hook called when all commands have finished (either successful or not).
-     */
-    onFinish() {}
 };
--- a/src/flow-control/base-handler.js
+++ b/src/flow-control/base-handler.js
@@ -8,4 +8,9 @@ module.exports = class BaseHandler {
     handle(commands) {
         return commands;
     }
+
+    /**
+     * A hook called when all commands have finished (either successful or not).
+     */
+    onFinish() {}
 };
--- a/src/flow-control/base-handler.js
+++ b/src/flow-control/base-handler.js
@@ -1,0 +1,2 @@
+module.exports = class BaseHandler {
+};
--- a/src/flow-control/input-handler.js
+++ b/src/flow-control/input-handler.js
@@ -2,8 +2,9 @@ const Rx = require('rxjs');
 const { map } = require('rxjs/operators');
 
 const defaults = require('../defaults');
+const BaseHandler = require('./base-handler');
 
-module.exports = class InputHandler {
+module.exports = class InputHandler extends BaseHandler {
     constructor({ defaultInputTarget, inputStream, logger }) {
         this.defaultInputTarget = defaultInputTarget || defaults.defaultInputTarget;
         this.inputStream = inputStream;
--- a/src/flow-control/input-handler.js
+++ b/src/flow-control/input-handler.js
@@ -5,11 +5,12 @@ const defaults = require('../defaults');
 const BaseHandler = require('./base-handler');
 
 module.exports = class InputHandler extends BaseHandler {
-    constructor({ defaultInputTarget, inputStream, logger }) {
+    constructor({ defaultInputTarget, inputStream, pauseInputStreamOnFinish, logger }) {
         super({ logger });
 
         this.defaultInputTarget = defaultInputTarget || defaults.defaultInputTarget;
         this.inputStream = inputStream;
+        this.pauseInputStreamOnFinish = pauseInputStreamOnFinish !== false;
     }
 
     handle(commands) {
@@ -40,7 +41,7 @@ module.exports = class InputHandler extends BaseHandler {
     }
 
     onFinish() {
-        if (this.inputStream) {
+        if (this.inputStream && this.pauseInputStreamOnFinish) {
             // https://github.com/kimmobrunfeldt/concurrently/issues/252
             this.inputStream.pause();
         }
--- a/src/flow-control/input-handler.js
+++ b/src/flow-control/input-handler.js
@@ -6,9 +6,10 @@ const BaseHandler = require('./base-handler');
 
 module.exports = class InputHandler extends BaseHandler {
     constructor({ defaultInputTarget, inputStream, logger }) {
+        super({ logger });
+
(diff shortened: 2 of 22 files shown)


brandonchinn178:
@kimmobrunfeldt It seems like Node 8 reached EOL on Jan 1. Can we remove Node 8 from the CI pipeline? Apparently `Promise.finally()` was added in Node 10

gustavohenke:
Hey @brandonchinn178 👋 
I've maintaining the package more than Kimmo actually -- I've bumped the minimum version to 10 on master.
Can you rebase and try again?

src/flow-control/input-handler.js
```diff
@@ -36,4 +39,11 @@ module.exports = class InputHandler {
 
         return commands;
     }
+
+    onFinish() {
+        if (this.inputStream && this.pauseInputStreamOnFinish) {
+            // https://github.com/kimmobrunfeldt/concurrently/issues/252
+            this.inputStream.pause();
```
gustavohenke, line 46: Rather than pausing, WDYT if this called `.unref()` instead?
We can then get rid of one the `pauseInputStreamOnFinish` option.

src/flow-control/base-handler.js
```diff
@@ -0,0 +1,16 @@
...
+    }
+
+    handle(commands) {
+        return commands;
+    }
+
+    /**
+     * A hook called when all commands have finished (either successful or not).
+     */
+    onFinish() {}
```
gustavohenke, line 15: WDYT if `handle` returned an optional cleanup function, along with the mapped commands?
Having an `onFinish` method kinda implies that the instance is single use.

gustavohenke (commented):
Hey! I'm sorry that I took this long to look at your PR.
Your analysis is very good, and it may be unveiling the reason for some other issues reported in that the processes hang forever!

src/flow-control/input-handler.js
brandonchinn178: `unref` is only a function on `net.Socket`, but not readable Stream, so it doesn't work if `stdin` is a file.
```
$ concurrently ... < foo
(node:51227) UnhandledPromiseRejectionWarning: TypeError: process.stdin.unref is not a function
...
```
We could update the docs to say "`inputStream` is either `net.Socket` or readable Stream" and then have this clean up check if the `unref` function exists and calls that if it does, but I think that adds too much complexity. Thoughts?

src/flow-control/base-handler.js
brandonchinn178: Personally, I like this better:
* It keeps `handle` as a pure transformation `Commands => Commands`
* It makes the `reduce` logic a bit more complicated in the main function
* It might also be a breaking change, if any third-party APIs write their own handlers
    * The only way to not make it a breaking change would be to allow `handle` to return either `Commands` or `{ commands, cleanup }`, which is additional complexity for the caller to handle

But it's your call

src/flow-control/input-handler.js
gustavohenke: Ah yes, of course it couldn't be this simple 🤦 OK. Need to think a bit more.

src/flow-control/base-handler.js
gustavohenke: For this one I think it's better to introduce a breaking change.
As you correctly stated, third parties can pass their own handlers - and it could become a pain point to manage state across multiple runs of the same instance.

Some other uses I can see for this cleanup function:
- Removing the RxJS subscription (which also removes the event from the `inputStream`, though doesn't make the process exit if it's `stdin`)
- Removing listeners on signals in the `KillOnSignal` handler (which can also prevent the process from exiting)

brandonchinn178: Yeah, that makes sense. I can update the handle api if you want. Are there any docs that I'd need to update as well?

brandonchinn178:
@gustavohenke any progress on this?

src/flow-control/input-handler.js
gustavohenke: Hmm, I think we could be smart and pause the input stream if it has no more listeners and it either
- was paused before; or
- was not flowing before.

Although, since I haven't been able to give much attention to concurrently, I'd say just introducing the option is probably best.

WDYT?

src/flow-control/base-handler.js
gustavohenke: Nope, no docs on creating custom handlers to this date 😞

src/flow-control/input-handler.js
brandonchinn178: It's been so long since I worked on this PR, honestly, I can't provide any more insights. (it seems like you are similarly distracted as well) If you're suggesting merging the PR as-is, that would be preferable by me. Are there any more blocking concerns you'd like me to address first?

gustavohenke: Just the handler interface, as per https://github.com/open-cli-tools/concurrently/pull/253#discussion_r637614525

brandonchinn178:
@gustavohenke How's that?

gustavohenke (approved):
Looks very good to me! Thanks for doing this 🏆

PR #266, opened 2021-03-10
fix: cwd option not overwriting per command if specified in programmatic usage

zoton2:
Noticed this while using this library for the first time, so sorry if I missed something. The recently added `cwd` option for the programmatic usage works but the override for each command doesn't, it was just silently dropped. This should fix that issue. I also noticed there were no tests that would pick up on it, so added one, although I'm not well versed on writing tests so it may need some modifications.

### DIFF
--- a/src/concurrently.js
+++ b/src/concurrently.js
@@ -69,6 +69,7 @@ function mapToCommandInfo(command) {
         name: command.name || '',
         prefixColor: command.prefixColor || '',
         env: command.env || {},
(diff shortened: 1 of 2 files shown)


moshest:
@gustavohenke can you please merge this? This is a hot fix.

gustavohenke (approved):
🙌 thanks!

gustavohenke:
Published in v6.0.1!

PR #269, opened 2021-04-07
Fixed input handler when input contains a colon

PatrickSachs:
Hey there,

thanks for concurrently, we are really enjoying using it our current project! Saves everyone from having to run dozens of scrips manually when running it.

As of now --handle-input does not like `:` colons (e.g. `2:MyProject.Module.func(a: 1)`), always redirecting it to the default input target instead of the one specified.

This is due to the `limit` option on the `String.split` function simply counter intuitively omits all matches beyond the limit. (`'a:b:c'.split(':', 2) == ['a', 'b']`)

This PR fixes this behavior.


gustavohenke (approved):
Simple change, thank you!

gustavohenke:
Published as v6.0.2.

Issue #186, opened 2019-04-18
Color changes for wild cards


codeedog:
Love this tool. When I tried the following (`package.json` has two `start-` scripts):

    concurrently -c cyan,yellow "npm:start-*"

the output of both subtasks were displayed with cyan rather than the first being colored cyan and the second being colored yellow. Instructions indicate that when color count is less than command count, the last color listed is used. I suspect whatever does the color logging sees only the command line argument count (a single command) and doesn't realize its a wildcard command so doesn't look for the actual number of commands running. Or, the wildcard processing code doesn't update some command count global. Or, the wildcard processing code is somehow treated as only one process.

In any event, it'd be nice to have wildcard initiated commands individually colored.

My work around is to just explicitly start the commands by name and I still get the coloring. This works:

    concurrently --c yellow,cyan npm:start-ng npm:start-express"

In addition, to explore some edge cases, I wrote a simple shell script that prints to the terminal and sleeps (see below). When I run it in with the wildcard, the wildcard logging is treated as a single process. Examples:

    concurrently -k -c cyan,yellow,green "bash foo.sh" "npm:start-*"
    concurrently -k -c cyan,yellow,green "npm:start-*" "bash foo.sh"

The first line outputs the shell logging in cyan and the two wildcards in yellow.
The second line outputs the two wildcards in cyan and the shell logging in yellow.

foo.sh
```
echo "hello"
sleep 1
echo "slept 1"
sleep 5
echo "slept 5"
sleep 60
echo "slept 60"
```

daniloraisi:
I was analyzing the code, and it seems that the problem is the way concurrently prefix colors.

As of this line:
https://github.com/open-cli-tools/concurrently/blob/0bc490fa98b302488b6fa9cf25d5cea239d38321/bin/concurrently.js#L153

It is considering only the first command to apply the color instead of applying each one when using a wildcard.

The command parser needs a way to apply prefix color when splitting npm wildcards.

@milahu, I can work on a PR to fix this.

gustavohenke:
Fix out in v6.3.0.

Issue #234, opened 2020-07-07
is it possible to have colors with npm wildcards command?


anisg:
Is it possible to add colors to npm wildcards commands?

I can have colors without using wildcards, eg:
```
concurrently --names 'api,front' -c 'bgBlue.bold,bgMagenta.bold' 'echo api' 'echo front' 
```

But when I use npm wildcards there is no colors: e.g in package.json
```
"scripts": {
    "dev-api": "echo api",
    "dev-front": "echo front",
    "dev": "concurrently  -c 'bgBlue.bold,bgMagenta.bold' npm:dev-*"
  },
```

milahu:
duplicate of #210

gustavohenke:
Fix out in v6.3.0.

Issue #210, opened 2019-12-25
Prefix-colors does not work with wildcard command.


tandat2209:
Using `prefix-colors` with wildcard does not work.

Example : `concurrently "npm:watch-*" -c "red,green,blue" ` => Only show red

milahu:
duplicate of #186

gustavohenke:
Fix out in v6.3.0.

PR #286, opened 2021-09-23
Fix colors option when using wildcard commands

daniloraisi:
This PR fix Issues #186, #210, and #234

Now when executing the command
```bash
concurrently -c "cyan,yellow" "npm:start-*"
```
will produce the same result as
```bash
concurrently -c "cyan,yellow" "npm:start-ng" "npm:start-server"
```

When mixing commands like
```bash
concurrently -k -c cyan,yellow,green "bash foo.sh" "npm:start-*"
concurrently -k -c cyan,yellow,green "npm:start-*" "bash foo.sh"
```

~~Will produce `bash foo.sh` to print as **cyan** or **yellow** respectively since the prefixColor attribution for plain commands use original index args passed to `concurrently` script.~~

Will produce `bash foo.sh` to print as **cyan** or **green**, since the attribution of prefixColor will happen after all commands are analyzed and prepared.


src/concurrently.js
```diff
@@ -79,6 +86,7 @@ function mapToCommandInfo(command) {
         command: command.command || command,
         name: command.name || '',
         prefixColor: command.prefixColor || '',
```
gustavohenke, line 88: we can get rid of this line now (and all its wiring)

gustavohenke (changes requested):
Hey, thanks for taking this on!

> concurrently -k -c cyan,yellow,green "npm:start-*" "bash foo.sh"
> Will produce bash foo.sh to print as cyan or green,

Did you mean yellow or green? Or am I missing something in your fix..?

daniloraisi:
> Hey, thanks for taking this on!
> 
> > concurrently -k -c cyan,yellow,green "npm:start-*" "bash foo.sh"
> > Will produce bash foo.sh to print as cyan or green,
> 
> Did you mean yellow or green? Or am I missing something in your fix..?

@gustavohenke when using this command `concurrently -k -c cyan,yellow,green "npm:start-*" "bash foo.sh"` the right thing is to color the output for `bash foo.sh` with the last color available.

Suppose you have only one script that matches `npm:start-*`. It will use cyan, so yellow will be the color used for the other command. For example, if you have two scripts that match `npm:start-*`, the last available color will be green, so `bash foo.sh` will be green.

src/concurrently.js
daniloraisi: I did that. Thanks!

gustavohenke (approved):
@daniloraisi I've pushed a couple of changes to your branch to make prefixes specified per command be preferred over those specified in the `prefixColors` option, plus added some docs and test.

We should be good to ship this now 🙌

gustavohenke:
Fix out in v6.3.0.

PR #287, opened 2021-10-02
Typescript rewrite

gustavohenke:


gustavohenke:
I'm going to take more of a hybrid approach instead of going full on.
In fact, `bin` is already converted to TS!

PR #289, opened 2021-10-04
Add support for options in environment variables

aidansteele:
This uses yargs' support for providing option values via the command line:  https://github.com/yargs/yargs/blob/main/docs/api.md#envprefix

My use case is that in CI scenarios I want raw output, but locally I want prefixes. I _could_ duplicate all the scripts in my package.json: one with `concurrently -r`, one without - but I feel that is suboptimal and could lead to mistakes when devs forget to keep them in sync.

With this change, my CI script can now look like:

```
set -eux
export CONCURRENTLY_RAW=true 

npm run check
npm run validate
npm run build-all
# etc..
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,10 @@ Examples:
      $ concurrently --names "HTTP,WATCH" -c "bgBlue.bold,bgMagenta.bold"
      "http-server" "npm run watch"
 
+ - Configuring via environment variables with CONCURRENTLY_ prefix
+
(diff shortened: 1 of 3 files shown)


aidansteele:
Sure thing, I'll try add those today :)

gustavohenke (approved):
I've added the docs and skipped the tests for now, there's too much repetition in the binary spec file 😐

PR #295, opened 2021-11-16
Add timing functionality

### LINKED ISSUE #291: Add option to show timings for each command in cli and also return timings when used programmatically
It would be useful if the package tracked the start and end times of processes and/or the calculated duration in ms, then returning this as part of `ExitInfos` for programmatic use or printing to terminal when a timing option is enabled.

eliasm307:
Add timing functionality to resolve #291

### DIFF
--- a/bin/concurrently.js
+++ b/bin/concurrently.js
@@ -62,10 +62,10 @@ const args = yargs
             default: defaults.hide,
             type: 'string'
         },
-        'show-timings': {
+        'timings': {
             describe: 'Show timing information for all processes',
             type: 'boolean',
-            default: defaults.showTimings
+            default: defaults.timings
         },
 
         // Kill others
@@ -147,7 +147,7 @@ const args = yargs
                 'Can be either the index or the name of the process.'
         }
     })
-    .group(['m', 'n', 'name-separator', 'raw', 's', 'no-color', 'hide', 'show-timings'], 'General')
+    .group(['m', 'n', 'name-separator', 'raw', 's', 'no-color', 'hide', 'timings'], 'General')
     .group(['p', 'c', 'l', 't'], 'Prefix styling')
     .group(['i', 'default-input-target'], 'Input handling')
     .group(['k', 'kill-others-on-fail'], 'Killing other processes')
@@ -177,7 +177,7 @@ concurrently(args._.map((command, index) => ({
     restartTries: args.restartTries,
     successCondition: args.success,
     timestampFormat: args.timestampFormat,
-    showTimings: args.showTimings
+    timings: args.timings
 }).then(
     () => process.exit(0),
     () => process.exit(1)
--- a/bin/concurrently.js
+++ b/bin/concurrently.js
@@ -62,6 +62,11 @@ const args = yargs
             default: defaults.hide,
             type: 'string'
         },
+        'show-timings': {
+            describe: 'Show timing information for all processes',
+            type: 'boolean',
+            default: defaults.showTimings
+        },
 
         // Kill others
         'k': {
@@ -142,7 +147,7 @@ const args = yargs
                 'Can be either the index or the name of the process.'
         }
     })
-    .group(['m', 'n', 'name-separator', 'raw', 's', 'no-color', 'hide'], 'General')
+    .group(['m', 'n', 'name-separator', 'raw', 's', 'no-color', 'hide', 'show-timings'], 'General')
     .group(['p', 'c', 'l', 't'], 'Prefix styling')
     .group(['i', 'default-input-target'], 'Input handling')
     .group(['k', 'kill-others-on-fail'], 'Killing other processes')
@@ -172,6 +177,7 @@ concurrently(args._.map((command, index) => ({
     restartTries: args.restartTries,
     successCondition: args.success,
     timestampFormat: args.timestampFormat,
+    showTimings: args.showTimings
 }).then(
     () => process.exit(0),
     () => process.exit(1)
--- a/src/completion-listener.js
+++ b/src/completion-listener.js
@@ -2,10 +2,9 @@ const Rx = require('rxjs');
 const { bufferCount, switchMap, take } = require('rxjs/operators');
 
 module.exports = class CompletionListener {
-    constructor({ successCondition, scheduler, showTimings}) {
+    constructor({ successCondition, scheduler }) {
         this.successCondition = successCondition;
         this.scheduler = scheduler;
-        this.showTimings = showTimings;
     }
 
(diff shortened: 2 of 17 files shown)


src/completion-listener.js
```diff
@@ -2,9 +2,10 @@ const Rx = require('rxjs');
 const { bufferCount, switchMap, take } = require('rxjs/operators');
 
 module.exports = class CompletionListener {
-    constructor({ successCondition, scheduler }) {
+    constructor({ successCondition, scheduler, showTimings}) {
```
gustavohenke, line 5: Why was this needed?

src/flow-control/log-timings.js
```diff
@@ -0,0 +1,57 @@
...
+
+module.exports = class LogTimings extends BaseHandler {
+    printExitInfoTimingTable(exitInfos) {
+        const exitInfoTable = _(exitInfos)
+            .map(({ command, timings, index, killed, exitCode }) => {
+                const readableDurationMs = (timings.endDate - timings.startDate).toLocaleString();
+                return {
+                    'call-index': index,
+                    name: command.name,
+                    duration: `~${readableDurationMs}ms`,
```
gustavohenke, line 14: Why the `~`? Is there any rounding somewhere I'm not spotting?

src/flow-control/log-timings.js
```diff
@@ -0,0 +1,57 @@
...
+const { bufferCount, take } = require('rxjs/operators');
+const BaseHandler = require('./base-handler');
+
+module.exports = class LogTimings extends BaseHandler {
+    printExitInfoTimingTable(exitInfos) {
+        const exitInfoTable = _(exitInfos)
+            .map(({ command, timings, index, killed, exitCode }) => {
+                const readableDurationMs = (timings.endDate - timings.startDate).toLocaleString();
+                return {
+                    'call-index': index,
```
gustavohenke, line 12: Is this meant to be different from `(index)`?

```
$ concurrently --show-timings "echo hi"
[0] echo hi started at 11/18/2021, 10:12:25 PM
[0] hi
[0] echo hi stopped at 11/18/2021, 10:12:25 PM after 18ms
[0] echo hi exited with code 0

Timings:
┌─────────┬────────────┬──────┬──────────┬───────────┬────────┬───────────┐
│ (index) │ call-index │ name │ duration │ exit-code │ killed │  command  │
├─────────┼────────────┼──────┼──────────┼───────────┼────────┼───────────┤
│    0    │     0      │  ''  │ '~18ms'  │     0     │ false  │ 'echo hi' │
└─────────┴────────────┴──────┴──────────┴───────────┴────────┴───────────┘
```

src/flow-control/log-timings.js
```diff
@@ -0,0 +1,57 @@
...
+                    name: command.name,
+                    duration: `~${readableDurationMs}ms`,
+                    'exit-code': exitCode,
```
gustavohenke, line 15: For some slightly added niceness 😊 

```suggestion
                    'exit code': exitCode,
```

src/flow-control/log-timings.js
```diff
@@ -0,0 +1,57 @@
...
+                    killed,
+                    command: command.command,
+                };
+            })
+            .sortBy('duration')
```
gustavohenke, line 20: Interesting, why sort by duration?

bin/concurrently.js
```diff
@@ -62,6 +62,11 @@ const args = yargs
             default: defaults.hide,
             type: 'string'
         },
+        'show-timings': {
```
gustavohenke, line 65: What do you think if this was only `--timings`?

src/flow-control/log-timings.js
```diff
@@ -0,0 +1,57 @@
...
+
+    handle(commands) {
+        if (!this.logger) { return {commands}; }
+
+        // individual process timings
+        commands.forEach(command => {
+            command.timer.subscribe( {
+                next: ({startDate, endDate}) => {
+                    if (!endDate) {
+                        this.logger.logCommandEvent( `${ command.command } started at ${ startDate.toLocaleString() }`, command );
```
gustavohenke, line 36: Wondering if we should use the format specified in `--timestamp-format` instead (hint, date-fns is a dependency!).

Might make _part_ of the test simpler/more stable on diverse machines due to all those regular expressions...

gustavohenke (commented):
whoaa this is super nice! I've added a bunch of thoughts, let me know what you think please.

There are some styling inconsistencies (but these aren't defined anywhere!), so I'll fix them up when we've agreed on the discussion below.

src/flow-control/log-timings.js
eliasm307: its just because the duration returned from in the exit info is more accurate, down to the nanosecond, and the two durations might not match but I guess adding this tilde just adds noise to the logs, I'll remove it

src/completion-listener.js
eliasm307: its a leftover from an old idea, will remove

src/flow-control/log-timings.js
eliasm307: so the `(index)` is the index in the sorted array, `call-index` is the index in terms of the order the processes were started, do you have an idea for a better name to make this clear? the `(index)` column is auto generated so cant change that easily unfortunately

src/flow-control/log-timings.js
eliasm307: I think if users are using the timing functionality they would be interested in the duration rankings ie knowing what was the fastest, slowest etc at a glance. Duration used to be a number but it wont sort properly as a string, I'll fix this

bin/concurrently.js
eliasm307: sounds good, I'll update

src/flow-control/log-timings.js
eliasm307: good idea, I'll look into it

src/flow-control/log-timings.js
eliasm307: maybe it doesnt matter here actually, I'll remove the call index and just leave the sorted index

eliasm307:
@gustavohenke I've applied your feedback, let me know if you see anything else

eliasm307:
@gustavohenke the test coverage being almost 100% was annoying me so I've added a few tests to cover the remaining lines

gustavohenke (approved):
Thanks! LGTM, I'm just implementing a logger-friendly `console.table` and will push directly to your branch.
We need to respect a bunch of other combinations users can have (raw mode, using a custom stdout, etc)

PR #296, opened 2021-11-19
Add auto colors functionality

### LINKED ISSUE #279: FR: Random Colors
is there anyway to just add the `--color` without defining colors for random colors?

eliasm307:
Resolves #279

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,7 +335,7 @@ For more details, visit https://github.com/open-cli-tools/concurrently
     Prefix colors specified per-command take precedence over this list.
   - `prefixLength`: how many characters to show when prefixing with `command`. Default: `10`
   - `raw`: whether raw mode should be used, meaning strictly process output will
-    be logged, without any prefixes, colouring or extra stuff.
+    be logged, without any prefixes, coloring or extra stuff.
   - `successCondition`: the condition to consider the run was successful.
     If `first`, only the first process to exit will make up the success of the run; if `last`, the last process that exits will determine whether the run succeeds.
     Anything else means all processes should exit successfully.
--- a/README.md
+++ b/README.md
@@ -188,8 +188,9 @@ Prefix styling
                           - Available modifiers: reset, bold, dim, italic,
                           underline, inverse, hidden, strikethrough
                           - Available colors: black, red, green, yellow, blue,
-                          magenta, cyan, white, gray
-                          or any hex values for colors, eg #23de43
+                          magenta, cyan, white, gray,
+                          any hex values for colors (e.g. #23de43) or auto for
+                          an automatically picked color
                           - Available background colors: bgBlack, bgRed,
                           bgGreen, bgYellow, bgBlue, bgMagenta, bgCyan, bgWhite
                           See https://www.npmjs.com/package/chalk for more
@@ -227,6 +228,7 @@ Options:
   -h, --help         Show help                                         [boolean]
   -v, -V, --version  Show version number                               [boolean]
 
+
 Examples:
 
  - Output nothing more than stdout+stderr of child processes
@@ -245,6 +247,14 @@ Examples:
 
      $ concurrently --names "HTTP,WATCH" -c "bgBlue.bold,bgMagenta.bold"
      "http-server" "npm run watch"
+
+ - Auto varying colored prefixes
+
+     $ concurrently -c "auto" "npm run watch" "http-server"
+
+ - Mixing auto and manual colored prefixes
+
+     $ concurrently -c "red,auto" "npm run watch" "http-server" "echo hello"
 
  - Configuring via environment variables with CONCURRENTLY_ prefix
 
@@ -320,8 +330,8 @@ For more details, visit https://github.com/open-cli-tools/concurrently
   - `prefix`: the prefix type to use when logging processes output.
     Possible values: `index`, `pid`, `time`, `command`, `name`, `none`, or a template (eg `[{time} process: {pid}]`).
     Default: the name of the process, or its index if no name is set.
-  - `prefixColors`: a list of colors as supported by [chalk](https://www.npmjs.com/package/chalk) or `auto` to automatically pick a color.
-    If concurrently would run more commands than there are colors, the last color is repeated, unless if the last colour value is `auto` which means following colors are automatically picked to vary.
+  - `prefixColors`: a list of colors as supported by [chalk](https://www.npmjs.com/package/chalk) or `auto` for an automatically picked color.
+    If concurrently would run more commands than there are colors, the last color is repeated, unless if the last color value is `auto` which means following colors are automatically picked to vary.
     Prefix colors specified per-command take precedence over this list.
   - `prefixLength`: how many characters to show when prefixing with `command`. Default: `10`
   - `raw`: whether raw mode should be used, meaning strictly process output will
--- a/README.md
+++ b/README.md
@@ -227,7 +227,6 @@ Options:
   -h, --help         Show help                                         [boolean]
   -v, -V, --version  Show version number                               [boolean]
 
-
 Examples:
 
  - Output nothing more than stdout+stderr of child processes
@@ -321,10 +320,9 @@ For more details, visit https://github.com/open-cli-tools/concurrently
   - `prefix`: the prefix type to use when logging processes output.
     Possible values: `index`, `pid`, `time`, `command`, `name`, `none`, or a template (eg `[{time} process: {pid}]`).
     Default: the name of the process, or its index if no name is set.
-  - `prefixColors`: a list of colors as supported by [chalk](https://www.npmjs.com/package/chalk).
-    If concurrently would run more commands than there are colors, the last color is repeated.
+  - `prefixColors`: a list of colors as supported by [chalk](https://www.npmjs.com/package/chalk) or `auto` to automatically pick a color.
+    If concurrently would run more commands than there are colors, the last color is repeated, unless if the last colour value is `auto` which means following colors are automatically picked to vary.
     Prefix colors specified per-command take precedence over this list.
-  - `colors`: let colours be selected to vary automatically where not explicitly defined
   - `prefixLength`: how many characters to show when prefixing with `command`. Default: `10`
   - `raw`: whether raw mode should be used, meaning strictly process output will
     be logged, without any prefixes, colouring or extra stuff.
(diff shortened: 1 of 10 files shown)


clement-at-beekast:
It would be great to merge this feature as almost everybody don't care about which colors to show in his project nor update it =)

paescuj:
@eliasm307 Thanks for this pull request! I've now rebased it on the latest state of the main branch. I've only made the necessary changes required for TypeScript & ESM and otherwise kept the original logic the same. Let's see if @gustavohenke and / or me can find the time to review this pull request soon.

README.md
```diff
@@ -323,6 +323,7 @@ concurrently can be used programmatically by using the API documented below:
   - `prefixColors`: a list of colors as supported by [chalk](https://www.npmjs.com/package/chalk).
     If concurrently would run more commands than there are colors, the last color is repeated.
     Prefix colors specified per-command take precedence over this list.
+  - `colors`: let colours be selected to vary automatically where not explicitly defined
```
gustavohenke, line 326: For consistency with the option name itself

```suggestion
  - `colors`: let colors be selected to vary automatically where not explicitly defined
```

src/prefix-color-selector.ts
```diff
@@ -0,0 +1,77 @@
...
+                chalk.bgWhite,
+                chalk.bgCyan,
+                chalk.bgGreen,
+                chalk.bgYellow,
+                chalk.bgRed,
+                chalk.bgGreenBright,
+                chalk.bgGrey,
+                chalk.blueBright,
+            ]
+                // Filter out duplicates
```
gustavohenke, line 28: is this necessary because chalk will return same colour depending on terminal support?

src/prefix-color-selector.ts
```diff
@@ -0,0 +1,77 @@
...
+    constructor(private readonly prefixColors?: string[], private readonly color?: boolean) {}
+
+    getNextColor(index?: number) {
+        const cannotSelectColor = !this.prefixColors?.length && !this.color;
+        if (cannotSelectColor) {
+            return '';
+        }
+
+        const userDefinedColorForCurrentCommand =
+            this.prefixColors && typeof index === 'number' && this.prefixColors[index];
```
gustavohenke, line 45: nitpick: this class is always invoked from within concurrently, so it's safe to assume typescript is having our back (assuming you apply the other suggestion)

```suggestion
            this.prefixColors && this.prefixColors[index];
```

src/prefix-color-selector.ts
```diff
@@ -0,0 +1,77 @@
...
+                .filter((chalkColor, index, arr) => {
+                    return arr.indexOf(chalkColor) === index;
+                })
+                .map(chalkColor => chalkColor.bold.toString())
+        );
+    }
+
+    constructor(private readonly prefixColors?: string[], private readonly color?: boolean) {}
+
+    getNextColor(index?: number) {
```
gustavohenke, line 38:
```suggestion
    getNextColor(index: number): string {
```

It's always being called with a value, so no need to make it optional

gustavohenke: Can we also add a jsdoc to this method?

src/prefix-color-selector.ts
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        // Auto selection requested and no user preference defined, select next auto color
+        if (!this.autoColors || !this.autoColors.length) {
+            this.refillAutoColors();
+        }
+
+        // Prevent consecutive colors from being the same
+        // (i.e. when transitioning from user colours to auto colours)
+        const nextColor = this.autoColors.shift();
```
gustavohenke, line 67: This method has a mix of stateful and stateless behaviour; you can specify an index that is within the bounds of `prefixColors`, and it'll always return that (stateless/idempotent).

But otherwise, the result depends on how many times `getNextColor` was called before (stateful/random), so `index` becomes meaningless.

I'd like to get rid of this ambiguity, so here are some suggestions:
1. Remove the `index` argument. Let `getNextColor` always return a "random" value depending on how many times it was called before.
2. Iterate up to `index` times generating the colour for each index in a predictable way and return it. E.g. calling `getNextColor(10); getNextColor(10)` always returns the same value. This can make the method `O(N)`, so some optimisation is required.

gustavohenke (commented):
Overall thought that doesn't one specific line:
What do you @eliasm307 and @paescuj think of `--prefix-colors` taking a special keyword `auto` that triggers this behaviour? So that this doesn't require implementing/specifying `--color`.

E.g. `concurrently -c auto,red,auto` npm:lint npm:test npm:build` generates an auto colour for `npm:lint`, red for `npm:test`, and another auto colour for `npm:build` (but excluding red)

eliasm307:
@gustavohenke thats an interesting suggestion, I think it could fit some use cases where a user wants partial control of the process colours.

However, for me the motivation for this was that I didn't really mind about the colours, as long as they are different. If other users also don't mind and they are trying to use this to spawn several processes then the command ends up with noise like "auto,auto,auto..." which isn't a good experience imo.

How would you feel about supporting both? So one option of all auto colours or being able to specify colours per process with an option to delegate?

src/prefix-color-selector.ts
eliasm307: The result also depends on the user colours provided to prevent repetition, which affects the order of the colours. It should be repeatable for the same initial conditions, but results will vary with different user inputs, which I think is acceptable in this case but please let me know if you feel strongly about it.

I will look into this to understand the code again, however looking at it after a while I don't think it's a good implementation. What I was trying to do is have the getNext colour method act as an infinite stream of colours but it looks odd giving it an index. 

What do you think about using a generator function for this? Would you allow that in the code base?

src/prefix-color-selector.ts
eliasm307: It's been a while but I feel like this was probably to guard against human error, I might have been imaging the list of colours growing over time so it's difficult to manually see duplicates

src/prefix-color-selector.ts
eliasm307: I agree, not sure what I was thinking here 😂

README.md
paescuj: Also I think we should name the option more precise, so that it is clear that only the prefixes are affected by it.

src/prefix-color-selector.ts
paescuj: Makes sense, IMO 👍 We could use a [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) for this.

src/prefix-color-selector.ts
paescuj: Actually, with the current implementation, the method is called without an index 
[line 69](https://github.com/open-cli-tools/concurrently/pull/296/files#diff-79dc04acf6a15308fee9ffe81596ff6718492b006c7ecec874448c9f64468e4aR69).

gustavohenke:
> command ends up with noise like "auto,auto,auto..." which isn't a good experience imo.

We could rely on the current behaviour of the last _input_ colour getting repeated when there are less colours than commands. So a single `auto` would do it if you had 20 commands.

Also sorry for reviewing this harshly, I didn't notice it was from a bunch of months back! 😱

paescuj:
> We could rely on the current behaviour of the last _input_ colour getting repeated when there are less colours than commands. So a single `auto` would do it if you had 20 commands.

I've already considered this too.
The advantage to this would be that we don't need to introduce another option.
The disadvantage in my opinion is that it might not be very intuitive at first glance. For example if I want to have auto colors for all commands except the second one, then I had to use `auto,red,auto`. I think this is okay, but we would need to document it clearly and provide a few examples.

src/prefix-color-selector.ts
eliasm307: updated

src/prefix-color-selector.ts
eliasm307: implementation changed, also its great the code base is in TS now 👍🏾! when I did this initially it was still in JS

src/prefix-color-selector.ts
eliasm307: implementation changed to use a generator, awaiting review

README.md
eliasm307: following the feedback I've opted to include this functionality as part of the `prefixColors` option, however the auto color functionality is opt-in so: 
- if someone wants all auto colors they just do `concurrently -c auto` 
- if they want to specify some colors but want the rest to be auto they can do so with the same option e.g. `concurrently -c red,green,auto,red,auto`
- if they want the current behaviour of repeating the last color then they just need to omit the last `auto` e.g. `concurrently -c red,green,auto,red`

eliasm307: docs text tbc, what do you guys think?

src/prefix-color-selector.ts
eliasm307: updated

eliasm307:
> > command ends up with noise like "auto,auto,auto..." which isn't a good experience imo.
> 
> We could rely on the current behaviour of the last _input_ colour getting repeated when there are less colours than commands. So a single `auto` would do it if you had 20 commands.
> 
> Also sorry for reviewing this harshly, I didn't notice it was from a bunch of months back! 😱

Hey, I didnt think it was harsh, it was good feedback, thanks for taking the time to review it 🙂

eliasm307:
Example output:
image

paescuj (approved):
Thanks @eliasm307 for your great work here!
It looks good to me 👍 
I just did some small enhancements concerning docs & comments and a little cleanup.

paescuj:
What do you think @gustavohenke?

src/prefix-color-selector.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+            expect(actualSelectedColor).not.toBe(previousColor); // No consecutive colors
+            expect(actualSelectedColor).toBe(currentExpectedColor); // Expected color
+            return actualSelectedColor;
+        }, '');
+    });
+});
+
+describe('PrefixColorSelector#ACCEPTABLE_CONSOLE_COLORS', () => {
+    it('has more than 1 auto color defined', () => {
+        // ! Code assumes this always has more than one entry, so make sure
```
gustavohenke, line 169: make sure to...?

gustavohenke (approved):
🤦 I had reviewed just hadn't submitted

src/prefix-color-selector.spec.ts
paescuj: Thanks! This is comment is now clarified 👍

gustavohenke:
Published in v7.5.0!

PR #303, opened 2022-01-12
Update docs example `concurrently()` after v7.0.0

naxoc:
In v7.0.0 `concurrently()` behaves differently. This updates the README code example to reflect that.


gustavohenke (approved):
oh my, missed that, thanks for updating it!

PR #281, opened 2021-06-01
Added Success Conditon All But First

woody34:
fixes #280 

- proposed feature to allow new success condition `all-but-first`

PR #318, opened 2022-05-07
Add command-{name|index} and !command-{name|index} condition to --success

### LINKED ISSUE #280: [Feature Request] Add success condition `all but first`
## Description

I would like to add a third success condition that would allow you to spin up a service, complete a list of tasks in parallel, capture their exit status. Assuming all tasks exited successfully, exit application with success status.

## Use case

I currently use concurrently to spin up storybook and complete cypress tests against a component library. I would like to be able to also perform jest visual regression tests.

## Proposed Use Case

```
const waitForStorybook = 'npx wait-on http://localhost:6006/';
concurrently(
	[
		{ name: 'Storybook', command: 'npm:sb', prefixColor: 'bgMagenta.bold' },
		{ name: 'Cypress', command: `${waitForStorybook} && npm run test:cypress:run`, prefixColor: 'bgGreen.bold' },
		{ name: 'Jest', command: `${waitForStorybook} && npm run test:unit`, prefixColor: 'bgBlue.bold' },
	],
	{
		prefix: 'name',
		successCondition: 'all-but-first',
	},
)
	.catch(console.error);
```

I am currently nesting concurrently calls to achieve this behavior and it doesn't provide the best logs.

## Proposed Acceptance Criteria

1. run 3 tasks concurrently
2. once task 2 and 3 complete with success, exit app with success

gustavohenke:
Implements 2 (or 4?) new possible conditions for `--success` flag:
- `command-{name|index}`: command with the specified name or index must exit with 0;
- `!command-{name|index}`: all but the command with the specified name or index must exit with 0.

I think that it expanding the use of `!` to `!first` and `!last` could come next, but for now that's an advanced enough solution!

Closes #280.
Related: #281

### DIFF
--- a/src/completion-listener.ts
+++ b/src/completion-listener.ts
@@ -8,8 +8,10 @@ import { CloseEvent, Command } from './command';
  * - `first`: only the first specified command;
  * - `last`: only the last specified command;
  * - `all`: all commands.
+ * - `command-{name|index}`: only the command with the specified name or index.
+ * - `!command-{name|index}`: all commands but the one with the specified name or index.
  */
-export type SuccessCondition = 'first' | 'last' | 'all';
+export type SuccessCondition = 'first' | 'last' | 'all' | `command-${string|number}` | `!command-${string|number}`;
 
 /**
  * Provides logic to determine whether lists of commands ran successfully.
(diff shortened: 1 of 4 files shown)


paescuj (commented):
Hey @gustavohenke, the implementation looks good!

Basically, I like the solution with the `command-{name|index}` pattern. Just two thoughts from my side:
* The command _name_ is not unique (while the _index_ is), there can be several commands with the same name. It is important to understand that in this case, the command that has finished first (!) will be considered. Should we add a note for this?
* It may happen that the command with the specified name or index does not exist.
   In the case of `!command-` concurrently exits with code 0, in the case of `command-` concurrently exits with code 1. Should we cover this in a test case? Should we also add a note for this?

gustavohenke:
> there can be multiple commands with the same name

Addressed that. Now all commands with the same name must meet the same condition.
I've also amended `--success` help to use the plural.

Noteworthy: input handling also accepts a target command name, but it's also only sending it to the first matching command...

> It can happen that the command with the specified name or index does not exist.
> In the case of `!command-` concurrently exits with code 0, in the case of `command-` concurrently exits with code 1.
> Should we cover this in a test case? Should we also add a note for this?

Covered it in tests -- might be fine to get away without more docs... I'd love to be able to write more about each flag, but `--help` already is pretty long 😞

src/completion-listener.ts
```diff
@@ -8,8 +8,10 @@ import { CloseEvent, Command } from './command';
  * - `first`: only the first specified command;
  * - `last`: only the last specified command;
  * - `all`: all commands.
+ * - `command-{name|index}`: only the command with the specified name or index.
+ * - `!command-{name|index}`: all commands but the one with the specified name or index.
```
paescuj, line 12:
```suggestion
 * - `command-{name|index}`: only the commands with the specified name or index.
 * - `!command-{name|index}`: all commands but the ones with the specified name or index.
```

paescuj:
Great work, LGTM!

> Addressed that. Now all commands with the same name must meet the same condition.
> I've also amended `--success` help to use the plural.

Nice, that makes very much sense 👍 

> Noteworthy: input handling also accepts a target command name, but it's also only sending it to the first matching command...

Okay, might be something to address in the feature...

> Covered it in tests -- might be fine to get away without more docs... I'd love to be able to write more about each flag, but `--help` already is pretty long 😞

Fine for me, I think it's a rare case anyway 👍 

Yeah, it would be handy to have a different place to to document such details!

PR #325, opened 2022-05-20
Fix `command-` syntax capturing for `--success`

### LINKED ISSUE #324: `command-${string}`-formatted `--success` condition fails with dashes
This should use the exit code from the first command (`exit 1`), but doesn't because [the `successCondition` splits on `-`](https://github.com/open-cli-tools/concurrently/blob/706c4d6435cf4f1a3a92a533af0ced22c03e7865/src/completion-listener.ts#L54):
```sh
concurrently \
  --names 'this-one,that-one' \
  --success command-this-one \
  'exit 1' \
  'exit 0'
```

This version does work:
```sh
concurrently \
  --names 'this,that' \
  --success command-that \
  'exit 1' \
  'exit 0'
```

paescuj:
Fixes #324

### DIFF
--- a/src/completion-listener.ts
+++ b/src/completion-listener.ts
@@ -43,7 +43,10 @@ export class CompletionListener {
             return events[0].exitCode === 0;
         } else if (this.successCondition === 'last') {
             return events[events.length - 1].exitCode === 0;
-        } else if (!/^!?command-.+$/.test(this.successCondition)) {
+        }
+
+        const commandSyntaxMatch = this.successCondition.match(/^!?command-(.+)$/);
+        if (commandSyntaxMatch === null) {
             // If not a `command-` syntax, then it's an 'all' condition or it's treated as such.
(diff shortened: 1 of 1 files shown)


src/completion-listener.ts
```diff
@@ -43,15 +43,18 @@ export class CompletionListener {
             return events[0].exitCode === 0;
         } else if (this.successCondition === 'last') {
             return events[events.length - 1].exitCode === 0;
-        } else if (!/^!?command-.+$/.test(this.successCondition)) {
+        }
+
+        const commandSyntaxMatch = this.successCondition.match(/^!?command-(.+)$/);
+        if (commandSyntaxMatch === null) {
```
gustavohenke, line 49: nitpick, if you like the suggestion...

```suggestion
        if (commandSyntaxMatch == null) {
```

gustavohenke (approved):

PR #330, opened 2022-06-14
Fix TypeScript not able to resolve types when set to Node16

Baune8D:
Fix for this error when using TypeScript 4.7 set to Node16 module resolution.

image


gustavohenke (approved):
great spot, thanks!

gustavohenke:
Published as v7.2.2 ✅

PR #334, opened 2022-06-23
Code Style & Performance Optimizations

paescuj:
* Prefer object spread & slice for better readability and performance
* Disallow `console.log` statements via ESLint
* Speed up GitHub workflow by doing "Lint & Format" checks only once per run and using better NPM cache setup
* Speed up tests by using `swc` as transformer and for CLI tests
* Add some missing test cases
* Move Jest config to own file for better readability and validation
* Optimize NPM script for prettier


src/concurrently.spec.ts
```diff
@@ -2,6 +2,10 @@ import { ChildProcess, KillProcess, SpawnCommand } from './command';
 import { concurrently, ConcurrentlyCommandInput, ConcurrentlyOptions } from './concurrently';
 import { createFakeProcess, FakeCommand } from './fixtures/fake-command';
 import { FlowController } from './flow-control/flow-controller';
+import { Logger } from './logger';
+import { OutputWriter } from './output-writer';
+
+jest.mock('./output-writer');
```
gustavohenke, line 8: That's one thing I don't enjoy very much from Jest.

`jest.mock` couples the test to the implementation.
You usually don't/shouldn't care about what a function is doing internally, as long as its results are what you expected.

I think that a better way to target the output writer would be to see if something gets written to the output stream when a logger is specified in the options instead.
WDYT?

bin/concurrently.spec.ts
```diff
@@ -17,14 +13,17 @@ const createKillMessage = (prefix: string) =>
  * Returns observables for its combined stdout + stderr output, close events, pid, and stdin stream.
  */
 const run = (args: string) => {
-    // TODO: This should only be transpiled once. Tests become 2.5x slower doing it in every `it`.
-    const child = spawn('ts-node --transpile-only ./concurrently.ts ' + args, {
+    // TODO: Optimally, this should only be transpiled once,
+    //       e.g. bundle in `beforeAll` and then reuse here.
+    const child = spawn(`node -r @swc-node/register ./concurrently.ts ${args}`, {
```
gustavohenke, line 18: Nice, I don't know much about SWC but I hope that the bin tests are much faster now?

src/command-parser/expand-arguments.ts
```diff
@@ -37,8 +37,6 @@ export class ExpandArguments implements CommandParser {
             }
         );
 
-        return Object.assign({}, commandInfo, {
```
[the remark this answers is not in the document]
paescuj: Yes, I think so too 👍

bin/concurrently.spec.ts
paescuj: Yes, SWC is great for such use cases 👍 It now takes about half as long as before 🚀

src/concurrently.spec.ts
paescuj: I'm pretty new to Jest so I really appreciate your feedback here!

> That's one thing I don't enjoy very much from Jest.
> 
> `jest.mock` couples the test to the implementation. You usually don't/shouldn't care about what a function is doing internally, as long as its results are what you expected.

I fully agree with you...

> I think that a better way to target the output writer would be to see if something gets written to the output stream when a logger is specified in the options instead.
> WDYT?

That makes much more sense! I've now updated the test in #335.

PR #335, opened 2022-06-27
Enhance tests

paescuj:
* Enhance the test for the logger option in `concurrently` as suggested by @gustavohenke in https://github.com/open-cli-tools/concurrently/pull/334#discussion_r906512429
  * While working on this test, I realized that we should extend the corresponding if statement with a check for `outputStream`, since it doesn't make sense to pass the log to the `OutputWriter` if there is no output stream.
* Increase coverage on Node.js 12 (we get a score of 💯 with this) by ensuring the `logger` is always defined in the `LogTimings` class.
  For this I marked the `printExitInfoTimingTable` method as private... @gustavohenke Do you think this would be a breaking change?

### DIFF
--- a/src/concurrently.spec.ts
+++ b/src/concurrently.spec.ts
@@ -3,9 +3,8 @@ import { concurrently, ConcurrentlyCommandInput, ConcurrentlyOptions } from './c
 import { createFakeProcess, FakeCommand } from './fixtures/fake-command';
 import { FlowController } from './flow-control/flow-controller';
 import { Logger } from './logger';
-import { OutputWriter } from './output-writer';
-
-jest.mock('./output-writer');
+import { Writable } from 'stream';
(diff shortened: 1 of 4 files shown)


gustavohenke (approved):
> Do you think this would be a breaking change?

It's not really documented how it's meant to be used, so probably not?

I never really thought of exposing the flow controllers  for their internals, but to allow one to customise which ones they want when manually building their concurrently instance.

paescuj:
> > Do you think this would be a breaking change?
> 
> It's not really documented how it's meant to be used, so probably not?
> 
> I never really thought of exposing the flow controllers for their internals, but to allow one to customise which ones they want when manually building their concurrently instance.

That's how I see it too! Additionally, `LogTimings` is the only controller with another method besides `handle`. So IMO it makes sense to mark it private.

PR #337, opened 2022-07-08
Windows testing improvements

### LINKED ISSUE #277: Please fix tests to run on windows
Tests fail at the moment when running on windows. This makes it hard for others to contribute to the project.

gustavohenke:
Closes #277

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,7 +10,7 @@ insert_final_newline = true
 indent_style = space
 
 # 4 space indentation and max line length of 100 in JavaScript/TypeScript files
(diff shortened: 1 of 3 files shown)


paescuj:
It turns out the tests for Windows are working on GitHub because their Windows environment comes with the 'sleep' command preinstalled. However, this is not the case on a default Windows setup.
Therefore I decided to use an own implementation of the 'sleep' command in the tests. This makes sure we're testing exactly the same on all platforms.

To be merged after #342 (because of the top-level-await).

PR #342, opened 2022-08-03
Drop support for Node.js v12 & add support for Node.js v18

paescuj:
Node.js v12 is EOL, so we should move away from it and support Node.js v18 (in "Active" state) instead.

See https://nodejs.org/en/about/releases/.

This is a **breaking change** and must be released under a new major version!

@gustavohenke What do you think about moving to [pure ESM](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c)? This would also enable us to use newer versions of `chalk` and `supports-color` which are pure ESM as well.
Something else that we might consider is transpiling / bundling...

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "conc": "./dist/bin/concurrently.js"
   },
   "engines": {
-    "node": "^14.13.0 || >=16.0.0"
+    "node": ">=14.13.0"
   },
   "exports": {
     ".": {
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "conc": "./dist/bin/concurrently.js"
   },
   "engines": {
-    "node": "^12.20.0 || ^14.13.0 || >=16.0.0"
+    "node": "^14.13.0 || >=16.0.0"
   },
   "exports": {
     ".": {
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "eslint-plugin-simple-import-sort": "^8.0.0",
     "jest": "^28.1.3",
     "jest-create-mock-instance": "^2.0.0",
-    "lint-staged": "^13.0.3",
+    "lint-staged": "^13.0.4",
     "prettier": "^2.6.2",
     "simple-git-hooks": "^2.7.0",
     "string-argv": "^0.3.1",
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "@swc/jest": "^0.2.21",
     "@types/jest": "^28.1.8",
     "@types/lodash": "^4.14.178",
-    "@types/node": "^16.11.47",
+    "@types/node": "^14.18.23",
     "@types/shell-quote": "^1.7.1",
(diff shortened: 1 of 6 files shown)


gustavohenke:
> What do you think about moving to [pure ESM](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c)?

I like this idea _with caveats_.

Node.js did a really awful job with the CJS/ESM compatibility, and going full ESM means never being useable in CJS codebases again (last I checked, that was the case at least).

So my main point is: **can we still support CJS codebases via transpiling somehow?**

It will also mean updating all imports to have file extensions. It's a one-off, but it'll be massive anyway.

> enable us to use newer versions of chalk and supports-color

I'd be happy to find alternative packages that work as well as these if we can't find any other way

paescuj:
> So my main point is: **can we still support CJS codebases via transpiling somehow?**

I'm going to look into this as soon as I have time! 👍

paescuj:
> Node.js did a really awful job with the CJS/ESM compatibility, and going full ESM means never being useable in CJS codebases again (last I checked, that was the case at least).
> 
> So my main point is: **can we still support CJS codebases via transpiling somehow?**

As you wrote, it is not possible to use ESM in CJS context not even by transpiling (at least not without some dirty hacks). Let's put the "transformation into a pure ESM package" on hold for now and look at this again at a later date...

paescuj:
I came up with https://github.com/open-cli-tools/concurrently/pull/366 as an initial approach to create bundles for both ESM & CJS while depending on pure ESM packages (`chalk` & `supports-color`). I've tested it by importing it in another module (ESM as well as CJS) and it seems to work so far...

paescuj:
@gustavohenke How do you feel about merging this along with #354? We can then treat #366 separately and merge it at a later time.

gustavohenke (approved):
> How do you feel about merging this along with https://github.com/open-cli-tools/concurrently/pull/354?

Will revisit #354, but this one LGTM

paescuj:
> > How do you feel about merging this along with #354?
> 
> Will revisit #354, but this one LGTM

Nice! 💪 I've just included Node.js 19 in the test and did a small syntactic clean-up in the `package.json` and `tsconfig.json` files.

I'm now going to update #354.

package.json
```diff
@@ -10,7 +10,7 @@
     "conc": "./dist/bin/concurrently.js"
   },
   "engines": {
-    "node": "^12.20.0 || ^14.13.0 || >=16.0.0"
+    "node": ">=14.13.0"
```
gustavohenke, line 13: There was a reason that v15 was skipped altogether:
https://github.com/open-cli-tools/concurrently/blob/d41e78d75af2bd5d32043e0600071963f8610494/index.mjs#L7-L8

paescuj: Oh, thanks! Sorry, haven't thought about v15. Fixed in #392.

PR #346, opened 2022-08-10
Enhance testing

paescuj:
* Upgrade Jest to v28
* Lint tests with the help of [eslint-plugin-jest](https://github.com/jest-community/eslint-plugin-jest)
* Solve all lint problems
  * Refactor tests using a promise based approach instead of the "done" callback
  * Run "alias" tests with `it.each` instead of defining them twice
  * Move `expectLinesForProcessStartAndStop` & `expectLinesForTimingsTable` methods to `expect.extend`
  * Fix & re-enable 'is of success when a SIGINT is sent' test
* Update all lint / test related dependencies
  * Downgrade `@types/node` to current active LTS version (v16)
    Generally, it's advised to align types with the Node.js version - since we're targeting multiple Node.js version we pin the types to the current active LTS version
* Bundle 'concurrently' only once for all tests in `bin/concurrently.spec.ts` with the help of [esbuild](https://github.com/evanw/esbuild)
  * Results in faster execution of the tests 🚀
  * Makes it possible to run the commands without a shell and thus get a correct behavior for the 'SIGINT' test


src/command.spec.ts
```diff
@@ -49,179 +52,125 @@ describe('#start()', () => {
...
-                    throw new Error('Unexpected call count');
-            }
-            callCount++;
-        });
-
         command.start();
         process.emit('close', 0, null);
+
+        expect(observerSpy.getValueAt(0)).toEqual({ startDate, endDate: undefined });
+        expect(observerSpy.getValueAt(1)).toEqual({ startDate, endDate });
```
gustavohenke, line 84: This is nice! But it seems it assumes that the observers will emit synchronously... or did I miss anything?

src/completion-listener.spec.ts
```diff
@@ -45,7 +45,7 @@ describe('with default success condition set', () => {
 
         scheduler.flush();
 
-        expect(result).rejects.toEqual(expect.anything());
+        return expect(result).rejects.toEqual(expect.anything());
```
gustavohenke, line 48: 💯

src/command.spec.ts
paescuj: Oh, you're right! Good catch, thanks! I'm going to address this in the next days...

paescuj: So what do you think about this? => #369 😃

PR #354, opened 2022-08-11
Switch to pnpm

paescuj:
[pnpm](https://pnpm.io/) feels like a great alternative to `npm`!

~The only downside is that we could probably no longer use `np` (https://github.com/sindresorhus/np/issues/251) but I'd suggest to automate the release process with a GitHub workflow, anyway.~

~To be merged after #342 (will update the pull request to use pnpm v7 then)~


.github/actions/setup-pnpm-cache/action.yml
```diff
@@ -0,0 +1,22 @@
+name: Setup PNPM cache
+runs:
+  using: composite
+  steps:
+    - name: Install pnpm
+      uses: pnpm/action-setup@v2.2.2
+      with:
+        version: 6
```
gustavohenke, line 8: Should this go in the `package.json` `engines`?

I'm not across pnpm differences between v6 and v7. I just hope it's not like yarn v1 and v2...

gustavohenke (changes requested):
> The only downside is that we could probably no longer use np

I've never used, tbh. I think that's a tool that @kimmobrunfeldt likes and documented somewhere 😄

.github/actions/setup-pnpm-cache/action.yml
paescuj: Apparently, the way to go is to use "packageManager" field in package.json.

> I'm not across pnpm differences between v6 and v7. I just hope it's not like yarn v1 and v2...

I don't think it's such a massive change as with yarn. But we might want to wait for #342 before merging this so we can use v7 from the beginning.

package.json
```diff
@@ -33,6 +33,7 @@
     "report-coverage": "cat coverage/lcov.info | coveralls",
     "test": "jest"
   },
+  "packageManager": "pnpm@7.17.0",
```
gustavohenke, line 36: Seems like this is this for [corepack](https://nodejs.org/api/corepack.html)?

gustavohenke (approved):
LGTM I think, is there anything still pending here? 😄

package.json
paescuj: Yeah, [`pnpm/action-setup`](https://github.com/pnpm/action-setup) requires a version to be set either directly via the `version` option (see https://github.com/open-cli-tools/concurrently/pull/354#discussion_r950709085) or via the `packageManager` property in `package.json`.

If you're okay with this, I'm going to revert it and use the `version` option of the action instead, which allows us to specify a range (e.g. `7`) instead of a fixed version (like we have now, `7.17.0`).

paescuj:
> LGTM I think, is there anything still pending here? 😄

Only the open conversation... I'm going to rebase and merge it afterwards 🚀

package.json
gustavohenke: > allows us to specify a range (e.g. 7) instead of a fixed version

SGTM

paescuj: Nice, thanks! I'll update the PR in a few minutes 👍

gustavohenke:
🤔 it seems pnpm is a pickier than npm when running publish:

```
$ pnpm publish
 ERR_PNPM_GIT_NOT_UNCLEAN  Unclean working tree. Commit or stash changes first.

If you want to disable Git checks on publish, set the "git-checks" setting to "false", or run again with "--no-git-checks".

$ git status
On branch main
Your branch is up to date with 'origin/main'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	tst-npm/

nothing added to commit but untracked files present (use "git add" to track)
```

(though I might be on a possibly old release of pnpm, and this might have been a bug)

paescuj:
> 🤔 it seems pnpm is a pickier than npm when running publish:
> 
> ```
> $ pnpm publish
>  ERR_PNPM_GIT_NOT_UNCLEAN  Unclean working tree. Commit or stash changes first.
> 
> If you want to disable Git checks on publish, set the "git-checks" setting to "false", or run again with "--no-git-checks".
> 
> $ git status
> On branch main
> Your branch is up to date with 'origin/main'.
> 
> Untracked files:
>   (use "git add <file>..." to include in what will be committed)
> 	tst-npm/
> 
> nothing added to commit but untracked files present (use "git add" to track)
> ```
> 
> (though I might be on a possibly old release of pnpm, and this might have been a bug)

Hmm you mean because it's not really "unclean" since there's only an untracked file? In that case it might indeed be a bit pickier and also this might have been addressed in a newer pnpm release...

PR #358, opened 2022-08-17
feat: add shorthand `conc`

kecrily:
`concurrently` is too long, which makes it unfriendly to non-English users and tiresome to type, which discourages potential users.

It simplifies the multiple-entry command `&&`, but not in the command line name. Let's add a shorthand to simplify it. Use `conc` shorthand

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,12 +57,13 @@ npm install concurrently --save
 
 ## Usage
 
-`concurrently` has an alias `conc`. I use conc throughout the following, but you can always replace `conc` with `concurrently`
+> **Note**
+> The `concurrently` command is now also available under the shorthand alias `conc`.
 
 Remember to surround separate commands with quotes:
 
 ```bash
-conc "command1 arg" "command2 arg"
+concurrently "command1 arg" "command2 arg"
 ```
 
 Otherwise **concurrently** would try to run 4 separate commands:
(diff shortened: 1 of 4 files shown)


paescuj:
Thank you @kecrily for your contribution! I like this idea! 👍

ToDo's:
- [ ] Update the docs
- [ ] Verify all things are working with the alias (help output, etc.)

paescuj:
@kecrily Would you like to work on the ToDo's mentioned above?

kecrily:
I have updated the documentation and verified it.

paescuj:
@gustavohenke Do you think we need to adjust something else? Also before merging this we should make sure the chosen alias fits well in the long term. For example, `cncr` (official abbreviation of "concurrent") would be another possibility, although I think `conc` is easier to remember. Any other ideas?

gustavohenke (commented):
I'm on limited computer time for a couple more weeks, so one thing I'd like to know is whether any alias will conflict with other popular/native OS tools.

From what I managed to test, it doesn't seem like a non-dev Mac has either `conc` or `cncr`.
It'd be good to know if an alias would conflict with tools on Windows or Linux too.

Also fwiw, concurrently used to have a `concurrent` alias back before v4:
7383eca03ee463013a4a6e83cd99045b97fbc8c9

I'll let you decide this one @paescuj 🙂

paescuj:
Thanks for your feedback @gustavohenke!

Let's go with `conc` as it clearly shows that it belongs to `concurrently`.

> I'm on limited computer time for a couple more weeks, so one thing I'd like to know is whether any alias will conflict with other popular/native OS tools.

That's a good point!
`conc` doesn't seem to conflict with other tools neither on Linux (checked with Ubuntu) nor on Windows (checked in PS, cmd & Google) and macOS (terminal, Homebrew). Also couldn't find another module on NPM which provides a command named `conc`.

> Also fwiw, concurrently used to have a `concurrent` alias back before v4: [7383eca](https://github.com/open-cli-tools/concurrently/commit/7383eca03ee463013a4a6e83cd99045b97fbc8c9)

Interesting 😄

paescuj:
@gustavohenke Could you publish a new version as soon as you have time?

gustavohenke:
Published as v7.4.0!

PR #366, opened 2022-08-24
WIP: ESM and CJS bundle

paescuj:
Attempt to create bundles for both ESM & CJS while depending on pure ESM packages (`chalk` & `supports-color`).
I've tested this by importing it in another module (ESM as well as CJS) and it seems to work so far...

Based on #342.

Any review / feedback is greatly appreciated! ❤️


paescuj:
Closing this in favor of complete switch to ESM.

PR #369, opened 2022-09-22
Fix command async tests

paescuj:
See https://github.com/open-cli-tools/concurrently/pull/346#discussion_r972684798


gustavohenke (approved):
Feels slightly over engineered, but perhaps there's not many better ways

paescuj:
I get you...

paescuj:
I get you... Of cour

PR #372, opened 2022-10-12
Fix onFinish signature when using exactOptionalPropertyTypes

Baune8D:
Seeing this error:

image

When consuming `concurrently` in a project with `exactOptionalPropertyTypes` enabled.

PR #343, opened 2022-08-03
Replace deprecated RxJS code

paescuj:
To be ready for RxJS v8 and get rid of annoying deprecation warnings in IDE.

Not sure if the solutions proposed in this pull request are the best, so any review / feedback is greatly appreciated! ❤️

- [x] Review `Replace deprecated RXJS signatures/functions` - Are there better solutions?
- [x] Probably revert `Add eslint plugin to report usage of deprecated code`

PR #374, opened 2022-10-15
Enable TypeScript strict mode

gustavohenke:
As title says. Addressing https://github.com/open-cli-tools/concurrently/pull/372#issuecomment-1277604405 😉

### DIFF
--- a/bin/concurrently.spec.ts
+++ b/bin/concurrently.spec.ts
@@ -59,12 +59,10 @@ const run = (args: string, ctrlcWrapper?: boolean) => {
 
     const stdout = readline.createInterface({
         input: child.stdout,
-        output: null,
     });
 
     const stderr = readline.createInterface({
         input: child.stderr,
-        output: null,
     });
 
     const log = new Rx.Observable<string>((observer) => {
@@ -82,8 +80,8 @@ const run = (args: string, ctrlcWrapper?: boolean) => {
     });
 
     const exit = Rx.firstValueFrom(
-        Rx.fromEvent(child, 'exit').pipe(
-            map((exit: [number | null, NodeJS.Signals | null]) => {
+        Rx.fromEvent<[number | null, NodeJS.Signals | null]>(child, 'exit').pipe(
+            map((exit) => {
                 return {
                     /** The exit code if the child exited on its own. */
                     code: exit[0],
@@ -180,7 +178,7 @@ describe('exiting conditions', () => {
                     // Instruct the wrapper to send CTRL+C to its child
                     sendCtrlC(child.process);
                 } else {
-                    process.kill(child.pid, 'SIGINT');
+                    process.kill(Number(child.pid), 'SIGINT');
                 }
             }
         });
--- a/src/command.ts
+++ b/src/command.ts
@@ -92,7 +92,7 @@ export class Command implements CommandInfo {
     readonly command: string;
 
     /** @inheritdoc */
-    readonly prefixColor: string;
+    readonly prefixColor?: string;
 
     /** @inheritdoc */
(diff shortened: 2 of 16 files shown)


bin/concurrently.spec.ts
```diff
@@ -82,8 +80,8 @@ const run = (args: string, ctrlcWrapper?: boolean) => {
     });
 
     const exit = Rx.firstValueFrom(
-        Rx.fromEvent(child, 'exit').pipe(
-            map((exit: [number | null, NodeJS.Signals | null]) => {
+        Rx.fromEvent<[number | null, NodeJS.Signals | null]>(child, 'exit').pipe(
```
gustavohenke, line 83: Sorry, I realise this is more trouble for #343.
I think we could get away with a `resultSelector` arg that casts the value though

src/command.ts
```diff
@@ -92,7 +92,7 @@ export class Command implements CommandInfo {
     readonly command: string;
 
     /** @inheritdoc */
-    readonly prefixColor: string;
+    readonly prefixColor?: string;
```
gustavohenke, line 95: Actually matches the `CommandInfo` interface.

src/command.ts
```diff
@@ -112,8 +112,9 @@ export class Command implements CommandInfo {
     killed = false;
     exited = false;
 
+    /** @deprecated */
```
gustavohenke, line 115: Let's remove this in a future PR 🙂

src/completion-listener.ts
```diff
@@ -92,11 +92,15 @@ export class CompletionListener {
...
+                        ? this.emitWithScheduler(Rx.of(exitInfos))
+                        : this.emitWithScheduler(Rx.throwError(exitInfos))
                 ),
                 take(1)
             )
         );
     }
+
+    private emitWithScheduler<O>(input: Rx.Observable<O>): Rx.Observable<O> {
+        return this.scheduler ? input.pipe(Rx.observeOn(this.scheduler)) : input;
```
gustavohenke, line 104: So RxJS types are a bit finicky. When you don't have a scheduler, you can't pass the arg at all to functions that accept it.

src/concurrently.ts
```diff
@@ -30,7 +30,7 @@ const defaults: ConcurrentlyOptions = {
  * If value is a string, then that's the command's command line.
  * Fine grained options can be defined by using the object format.
  */
-export type ConcurrentlyCommandInput = string | Partial<CommandInfo>;
+export type ConcurrentlyCommandInput = string | ({ command: string } & Partial<CommandInfo>);
```
gustavohenke, line 33: bit weird to allow a command not to have a command...

gustavohenke:
I forgot to rebase before pushing.
One moment please

gustavohenke:
Let me know of any changes you think might be useful, @paescuj 🙂

bin/concurrently.spec.ts
paescuj: No problem! We can address this in #343. We probably have to rely on `map` though as `resultSelector` is deprecated: https://rxjs.dev/deprecations/resultSelector

src/command.ts
paescuj: Good idea 👍

src/completion-listener.ts
paescuj: I see 👍 Good solution!

src/concurrently.ts
paescuj: Nice 😄

paescuj:
Thanks for taking this on and sorry for my late reply! Looks solid 👍😃

PR #376, opened 2022-10-31
update install command to use `--save-dev`

DetachHead:
concurrently is a dev tool so it should be saved to `devDependencies`

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ npm install -g concurrently
 or if you are using it from npm scripts:
 
 ```bash
-npm install concurrently --save
+npm install concurrently --save-dev
 ```
 
 ## Usage


gustavohenke (commented):
It's arguable.

Using concurrently to spin up multiple build tools at once is pretty common, so it fits correctly the description of a dev tool.

On the other hand, you could have multiple services in a production server that need to be respawned when one of them dies.
I'm not saying that this is a good practise, but it's a valid use case for concurrently.

@paescuj opinions?

paescuj:
I'm in favor of this change for the following two reasons:
- I assume that in a majority of projects "concurrently" is indeed used as a development tool.
- It feels safer to propose the installation as `devDependency` by default.

> I'm not saying that this is a good practise, but it's a valid use case for concurrently.

I also think this is a valid use case, but let's leave it to the users to declare "concurrently" as a productive dependency in this case.

PR #385, opened 2022-11-18
Correct wildcard naming as described in README

### LINKED ISSUE #312: Wildcard option doesn't seem to follow README
Despite #212, it seems the wildcard option still (v7.x) doesn't get `name` right, 
> concurrently "npm:watch-*"
> 
> Equivalent to:
> concurrently -n js,css,node "npm run watch-js" "npm run watch-css" "npm run watch-node"

but gives now:
`concurrently -n watch-*js,watch-*css,watch-*node "npm run watch-js" "npm run watch-css" "npm run watch-node"`

Something I understood wrong or a recurring bug?

paescuj:
Closes #312

### DIFF
--- a/src/command-parser/expand-npm-wildcard.spec.ts
+++ b/src/command-parser/expand-npm-wildcard.spec.ts
@@ -83,6 +83,25 @@ for (const npmCmd of ['npm', 'yarn', 'pnpm']) {
             ]);
         });
 
+        it('uses wildcard match of script as command name', () => {
+            readPkg.mockReturnValue({
+                scripts: {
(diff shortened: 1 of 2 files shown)


gustavohenke (commented):
Great way of fixing this!
I had mentioned in the linked issue that it'd be a breaking change, but like this it doesn't seem like it needs to be! 🙂

paescuj:
> Great way of fixing this! I had mentioned in the linked issue that it'd be a breaking change, but like this it doesn't seem like it needs to be! 🙂

Thanks! Exactly, it's more like a fix 😄

PR #390, opened 2022-11-24
Fixed broken link in the readme

abstractpoint:
Link to "task automation with npm" pointed to 404, now pointing to an archive of James` post with the same content.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Like `npm run watch-js & npm run watch-less` but better.
 
 ## Why
 
-I like [task automation with npm](https://github.com/substack/blog/blob/master/npm_run.markdown)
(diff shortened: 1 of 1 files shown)


paescuj:
Thank you!
Do you know by any chance whether James is affiliated with Founding Stack? If not, I would rather not refer to this page, because the site also contains some marketing content and I'd like to keep this repo as neutral as possible. I hope you understand that!
Alternatively, we could refer to the archived version from Wayback Machine: http://web.archive.org/web/20220531064025/https://github.com/substack/blog/blob/master/npm_run.markdown.

paescuj:
Thanks for the swift adjustment!

PR #393, opened 2022-11-28
Support REPL languages that use colons

wyattades:
It's common for a server command to have a built-in REPL for debugging, e.g. a Node.js server with `debugger` or a Rails server with `binding.pry`. This PR adds support for writing most syntaxes into the concurrently input stream. Specifically, we need to allow colons that commonly appear in code, without it changing the target input stream.

e.g. the following Javascript should be a valid input:
```js
myFunc({ foo: "bar" })
```
Before this PR, this would fail with the error: `Unable to find command "myFunc({ foo", or it has no stdin open`

**Breaking changes:**
- drop support for `1::something` (so we can support `Ruby::Const::Syntax`). we'll still support `1: :something` and `1:Ruby::Const::Syntax`

---

Speaking of REPL, it'd be great to support REPL keystrokes like backspaces, arrow left, etc. but I'm not sure how to do that...

### DIFF
--- a/src/flow-control/input-handler.spec.ts
+++ b/src/flow-control/input-handler.spec.ts
@@ -55,11 +55,13 @@ it('forwards input stream to target index specified in input', () => {
 
     inputStream.write('1:something');
     inputStream.write(' 1:ignore_leading_whitespace');
+    inputStream.write('1:multi\nline\n');
 
     expect(commands[0].stdin?.write).not.toHaveBeenCalled();
-    expect(commands[1].stdin?.write).toHaveBeenCalledTimes(2);
+    expect(commands[1].stdin?.write).toHaveBeenCalledTimes(3);
     expect(commands[1].stdin?.write).toHaveBeenCalledWith('something');
     expect(commands[1].stdin?.write).toHaveBeenCalledWith('ignore_leading_whitespace');
+    expect(commands[1].stdin?.write).toHaveBeenCalledWith('multi\nline\n');
 });
 
 it('forwards input stream to target index specified in input when input contains colon', () => {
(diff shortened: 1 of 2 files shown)


paescuj:
Thank you @wyattades for your pull request! ❤️
Let me propose an alternative solution: If the part before the colon doesn't match an actual command index / name, the whole input is sent to the default target. This would also address the issues you've mentioned above and would be a bit simpler IMO. WDYT @wyattades, @gustavohenke?

> Speaking of REPL, it'd be great to support REPL keystrokes like backspaces, arrow left, etc. but I'm not sure how to do that...

Interesting! We could take a look at the [Node.js REPL API](https://nodejs.org/api/repl.html) or [@stdlib/repl](https://www.npmjs.com/package/@stdlib/repl) for this...

wyattades:
> Let me propose an alternative solution: If the part before the colon doesn't match an actual command index / name, the whole input is sent to the default target. This would also address the issues you've mentioned above and would be a bit simpler IMO.

That seems like a good idea, although I think using a simple Regexp would still be an improvement over `.split`. So something like this:

```js
const dataMatch = data.match(/^\s*([\w-]+):(.+)$/s);
let command;
if (dataMatch) {
  command = commandsMap.get(dataMatch[1]);
}
// if `command` does not exist, don't log an error, just fallback to defaultInputTarget
command ||= commandsMap.get(this.defaultInputTarget.toString());

// now log an error if `!command?.stdin`...
```

gustavohenke (commented):
Oh wow, I never thought of using REPLs with input handling 😄 

> If the part before the colon doesn't match an actual command index / name, the whole input is sent to the default target.

Yes, I like this idea.

> although I think using a simple Regexp would still be an improvement over .split. 

Why is that? Is it so that you can trim the target command in one go?

wyattades:
@gustavohenke a gentle reminder :)

wyattades:
Hey @paescuj @gustavohenke, can you help me get this released?

paescuj:
> Hey @paescuj @gustavohenke, can you help me get this released?

Thanks and sorry for the delay!

Issue #395, opened 2022-12-08
Integration tests never complete


paulsmithkc:
**Environment**
MacOS 13.0.1
Node 16.16.0
Typescript 4.9.4
Concurrently 7.6.0

**Describe the bug**
We are attempting to write some integration tests for our azure functions, however the function runtime restarts instead of exiting when the test are complete, so the process never finishes.

**To Reproduce**
1. Create a new function app
2. Create at least one http trigger
3. Write an integration test using `jest` which makes a http request against the trigger
4. Install the following packages: jest, ts-jest, concurrently, wait-on
5. Add the following scripts to `package.json`

```
"pretest": "tsc -p tsconfig.test.json",
"test": "concurrently -k -s last \"func start\" \"wait-on tcp:7071 && jest\"",
```

6. Execute the following command on the terminal: `npm run test`
7. Wait for process to complete.

**Expected behavior**
When the jest exits, concurrently should kill the azure function runtime, and terminate.

**Actual behavior**
1. Jest exits successfully.
2. Concurrently sends a kill signal to the azure function runtime.
3. The azure function runtime restarts itself and keeps running.
4. Integration tests never complete, because runtime is still active.

**Relevant Console Output**
```
[1] Ran all test suites.
[1] wait-on tcp:7071 && jest exited with code 0
--> Sending SIGTERM to other processes..
[0] [2022-12-08T16:06:34.042Z] Language Worker Process exited. Pid=64056.
[0] [2022-12-08T16:06:34.042Z] node exited with code 143 (0x8F). .
[0] [2022-12-08T16:06:34.280Z] Worker process started and initialized.
```

**Proposed Fix**
Add a command line argument to `concurrently` which will send a SIGKILL signal instead of a SIGTERM signal, to force the azure runtime to exit immediately.

gustavohenke:
Are you able to provide a self contained example at all?
Creating an azure function is way too much trouble.

Are there any docs about the behaviour you're seeing?

NicolasBissig:
I have the same problem with Linux, windows works. My repo is here: https://github.com/NicolasBissig/azure-functions-decorators/tree/feature/examples

how to recreate:
- clone repo with branch `feature/examples`
- install dependencies
- in directory `packages/examples` run `npm test`

What you should see:
```
[jest]  PASS  __test__/function-http-query.test.ts (7.714 s)
[jest]
[jest] Test Suites: 5 passed, 5 total
[jest] Tests:       7 passed, 7 total
[...]

PR #396, opened 2022-12-08
Send SIGKILL to other processes

paulsmithkc:
When the main process exits, force the other processes to exit with the SIGKILL signal. (Instead of SIGTERM.)

Fixes https://github.com/open-cli-tools/concurrently/issues/395


paulsmithkc:
@gustavohenke bump

gustavohenke (changes requested):
Just changing the signal like this will most likely break lots of people's workflows.
Let's discuss the solution over at #395 first.

paulsmithkc:
Rejected.

Submitted new PR: https://github.com/open-cli-tools/concurrently/pull/402

PR #402, opened 2023-01-30
add --kill-signal option

### LINKED ISSUE #395: Integration tests never complete

paulsmithkc:
Add a --kill-signal option to change what signal is sent to the other processes.

Fixes #395

### DIFF
--- a/bin/concurrently.ts
+++ b/bin/concurrently.ts
@@ -110,6 +110,11 @@ const args = yargs(argsBeforeSep)
             describe: 'Kill other processes if one exits with non zero status code.',
             type: 'boolean',
         },
+        'kill-signal': {
+            alias: 'ks',
+            describe: 'Signal to send to other processes if one exits or dies.',
+            type: 'string',
+        },
 
         // Prefix
         prefix: {
@@ -208,6 +213,7 @@ concurrently(
             : args.killOthersOnFail
             ? ['failure']
             : [],
+        killSignal: args.killSignal,
         maxProcesses: args.maxProcesses,
         raw: args.raw,
         hide: args.hide.split(','),
--- a/bin/concurrently.ts
+++ b/bin/concurrently.ts
@@ -112,8 +112,10 @@ const args = yargs(argsBeforeSep)
         },
         'kill-signal': {
             alias: 'ks',
-            describe: 'Signal to send to other processes if one exits or dies.',
+            describe:
+                'Signal to send to other processes if one exits or dies. (SIGTERM/SIGKILL, defaults to SIGTERM)',
             type: 'string',
+            default: defaults.killSignal,
         },
 
         // Prefix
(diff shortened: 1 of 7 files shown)


paulsmithkc:
@paescuj @gustavohenke bump

jessephelps:
This would be helpful for my issues as well!

bin/concurrently.ts
```diff
@@ -110,6 +110,11 @@ const args = yargs(argsBeforeSep)
             describe: 'Kill other processes if one exits with non zero status code.',
             type: 'boolean',
         },
+        'kill-signal': {
+            alias: 'ks',
+            describe: 'Signal to send to other processes if one exits or dies.',
```
gustavohenke, line 115: It'd be useful to list what the default is here. Take a look at `passthrough-arguments` and `timings` for some examples

src/flow-control/kill-others.spec.ts
```diff
@@ -23,6 +23,7 @@ const createWithConditions = (conditions: ProcessCloseCondition[]) =>
     new KillOthers({
         logger,
         conditions,
+        killSignal: undefined,
```
gustavohenke, line 26: We need to extend the test file to cover the arguments passed to the `command.kill()`, both when `undefined` and the custom one

gustavohenke (changes requested):

bin/concurrently.ts
paulsmithkc: Fixed.

src/flow-control/kill-others.spec.ts
paulsmithkc: Fixed.

gustavohenke (approved):
Thanks 🎉

gustavohenke:
Released in v8.0.0!

PR #410, opened 2023-03-29
Upgrade to pnpm 8

paescuj:


paescuj:
Postponed to EOL of Node.js 14 (2023-04-30)

danny007in:
i think better to use only official npm

paescuj:
> better to use only latest npm is there any special use of pnpm

What do you mean exactly?

danny007in:
> mean exactly?

Better to replace pnpm to latest npm

gustavohenke:
Hey @danny007in, why do you think so?

danny007in:
> Hey @danny007in, why do you think so?

Exactly that is my point, using stable npm package best option it also  links files from the global store.

paescuj:
> [...] it also links files from the global store.

Did I miss something here? I'm not aware of this. Could you share a link?

Besides that, I don't think we're going back to `npm` (at least for now).
It has been shown that `pnpm` works very well for here. In particular, it has helped to greatly reduce the duration of the tests.

danny007in:
> > [...] it also links files from the global store.
> 
> Did I miss something here? I'm not aware of this. Could you share a link?
> 
> Besides that, I don't think we're going back to `npm` (at least for now). It has been shown that `pnpm` works very well here. In particular, it has helped to greatly reduce the duration of the tests.

`npm v9` works as pnpm now.

https://stackoverflow.com/questions/74569635/npm-9-doesnt-copy-js-files-for-local-dependencies

gustavohenke (commented):
@paescuj did you not want to proceed with this?

PR #411, opened 2023-03-31
Allow setting `raw` per command when using the API

chbiel:
In our setup, one of the processes is a docker compose process.
For this process we want to have `raw` output, but for other processes we would still like to have prefixes and colors.

I also removed one unused property. If you don't want that I can also take that commit out.

Thanks in advance

### DIFF
--- a/src/command.ts
+++ b/src/command.ts
@@ -102,6 +102,9 @@ export class Command implements CommandInfo {
     /** @inheritdoc */
     readonly env: Record<string, unknown>;
 
+    /** @inheritdoc */
+    readonly cwd?: string;
+
     readonly close = new Rx.Subject<CloseEvent>();
     readonly error = new Rx.Subject<unknown>();
     readonly stdout = new Rx.Subject<Buffer>();
@@ -120,7 +123,7 @@ export class Command implements CommandInfo {
     }
 
     constructor(
-        { index, name, command, prefixColor, env }: CommandInfo & { index: number },
+        { index, name, command, prefixColor, env, cwd }: CommandInfo & { index: number },
         spawnOpts: SpawnOptions,
         spawn: SpawnCommand,
         killProcess: KillProcess
@@ -130,6 +133,7 @@ export class Command implements CommandInfo {
         this.command = command;
         this.prefixColor = prefixColor;
         this.env = env || {};
+        this.cwd = cwd;
         this.killProcess = killProcess;
         this.spawn = spawn;
         this.spawnOpts = spawnOpts;
--- a/src/command.ts
+++ b/src/command.ts
@@ -32,6 +32,11 @@ export interface CommandInfo {
      * Color to use on prefix of the command.
      */
     prefixColor?: string;
+
+    /**
+     * Output command in raw format
+     */
+    raw?: boolean;
 }
 
 export interface CloseEvent {
--- a/src/command.ts
+++ b/src/command.ts
@@ -34,7 +34,7 @@ export interface CommandInfo {
     prefixColor?: string;
 
     /**
-     * Output command in raw format
+     * Output command in raw format.
      */
     raw?: boolean;
 }
--- a/src/command.ts
+++ b/src/command.ts
@@ -97,9 +97,6 @@ export class Command implements CommandInfo {
     /** @inheritdoc */
     readonly env: Record<string, unknown>;
 
-    /** @inheritdoc */
-    readonly cwd?: string;
-
     readonly close = new Rx.Subject<CloseEvent>();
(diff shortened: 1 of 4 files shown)


chbiel:
I see the failing tests, but it's unclear for me, how this could be related to my changes. Do you have an idea?

paescuj:
Thanks for your contribution ❤️ 

> I see the failing tests, but it's unclear for me, how this could be related to my changes. Do you have an idea?

Seems to be just a timing issue, unfortunately happens sometimes (not related to your changes).

chbiel:
hi @paescuj, do you have an idea when the PR can be merged?

paescuj:
Will try to review this evening 👍

src/concurrently.spec.ts
```diff
@@ -253,6 +253,62 @@ it('uses overridden cwd option for each command if specified', () => {
...
+        })
+    );
+    expect(spawn).toHaveBeenCalledWith(
+        'kill',
+        expect.objectContaining({
+            env: expect.not.objectContaining({ foo: expect.anything() }),
+            stdio: 'inherit',
+        })
+    );
+});
```
paescuj, line 290: Can we simplify this a bit? E.g. just two commands. Right now it feels more like a test for `env`.

src/concurrently.ts
```diff
@@ -162,7 +162,7 @@ export function concurrently(
                     ...command,
                 },
                 getSpawnOpts({
-                    raw: options.raw,
+                    raw: command.raw !== undefined ? command.raw : options.raw,
```
paescuj, line 165:
```suggestion
                    raw: command.raw ?? options.raw,
```

paescuj (changes requested):

src/concurrently.spec.ts
chbiel: Sure, I can remove one of them. I used the same approach here like in `uses cwd from options for each command`-test.
As far as I understand, there are not many more option to distinguish the same command here in the tests.

chbiel: Commited a simplified version of the test

paescuj (changes requested):
> I also removed one unused property. If you don't want that I can also take that commit out.

Why do you think it's unused? At least it's  documented here: https://github.com/open-cli-tools/concurrently#closeevent

chbiel:
> > I also removed one unused property. If you don't want that I can also take that commit out.
> 
> Why do you think it's unused? At least it's documented here: https://github.com/open-cli-tools/concurrently#closeevent

Yes and it is still a valid parameter as part of the `CommandInfo` type, but there is no usage with the `Command` class (no read or write).
The `cwd` configuration is only used as part of the `SpawnOptions`.

chbiel:
hey @paescuj, any updates here? As said I did not remove the `cwd` completely, but only from the `Command` class were it is not used

src/command.ts
```diff
@@ -97,9 +102,6 @@ export class Command implements CommandInfo {
     /** @inheritdoc */
     readonly env: Record<string, unknown>;
 
-    /** @inheritdoc */
-    readonly cwd?: string;
-
     readonly close = new Rx.Subject<CloseEvent>();
```
paescuj, line 105:
```suggestion
    /** @inheritdoc */
    readonly cwd?: string;

    readonly close = new Rx.Subject<CloseEvent>();
```

src/command.ts
```diff
@@ -32,6 +32,11 @@ export interface CommandInfo {
      * Color to use on prefix of the command.
      */
     prefixColor?: string;
+
+    /**
+     * Output command in raw format
```
paescuj, line 37:
```suggestion
     * Output command in raw format.
```

src/command.ts
```diff
@@ -118,7 +120,7 @@ export class Command implements CommandInfo {
     }
 
     constructor(
-        { index, name, command, prefixColor, env, cwd }: CommandInfo & { index: number },
+        { index, name, command, prefixColor, env }: CommandInfo & { index: number },
```
paescuj, line 123:
```suggestion
        { index, name, command, prefixColor, env, cwd }: CommandInfo & { index: number },
```

src/command.ts
```diff
@@ -128,7 +130,6 @@ export class Command implements CommandInfo {
         this.command = command;
         this.prefixColor = prefixColor;
         this.env = env || {};
-        this.cwd = cwd;
         this.killProcess = killProcess;
```
paescuj, line 133:
```suggestion
        this.cwd = cwd;
        this.killProcess = killProcess;
```

paescuj (approved):

PR #422, opened 2023-04-29
fix: move `types` condition to the front

Andarist:
I moved `types` condition to the front. `package.json#exports` are **order-sensitive** - they are always matched from the top to the bottom. When a match is found then it should be used and no further matching should occur. 

Right now, the current setup works in TypeScript but it's considered a bug and it should not be relied upon, see the thread and the comment [here](https://github.com/microsoft/TypeScript/issues/50762#issuecomment-1528318260). For that reason, I would like to fix all popular packages that misconfigured their `exports` this way so the bug can be fixed in TypeScript.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,10 +14,10 @@
   },
   "exports": {
     ".": {
+      "types": "./dist/src/index.d.ts",
       "import": "./index.mjs",
       "require": "./index.js",
-      "default": "./index.js",
-      "types": "./dist/src/index.d.ts"
+      "default": "./index.js"
     },
     "./package.json": "./package.json"
   },


paescuj:
Thanks!

Andarist:
@paescuj note that this is still not 100% correct, see [arethetypeswrong](https://arethetypeswrong.github.io/?p=concurrently%408.0.1):
> Imports of "concurrently" under the node16 module resolution setting when the importing module is ESM (its extension is .mts or .mjs, or it has a .ts or .js extension and is in scope of a package.json that contains "type": "module") resolved to CJS types, but ESM implementations.

This PR fixed specifically the other error (`🐛 Used fallback condition`).

Andarist:
@paescuj would there be a chance to cut a release with this small fix?

gustavohenke:
Done @Andarist!

PR #431, opened 2023-06-24
Delay summary timings output until finished

### LINKED ISSUE #316: When using --timings with --group, timings are printed in the middle of the output
I encountered a little unexpected behavior while using `concurrently` today, and wanted to file an issue in case it seemed like something worth tackling!

If you use the `--timings` along with the `--group` option, the timings table is printed when the last command exits rather than at the end of the output. This can lead to the timings table getting printed in the middle of the stream rather than the end:

```shell
❯ ./node_modules/.bin/concurrently --group --timings "sleep 1 && exit" "exit"
[0] sleep 1 && exit started at 2022-04-24 19:28:03.242
[0] sleep 1 && exit stopped at 2022-04-24 19:28:04.252 after 1,010ms
[0] sleep 1 && exit exited with code 0
--> Timings:
--> ┌──────┬──────────┬───────────┬────────┬─────────────────┐
--> │ name │ duration │ exit code │ killed │ command         │
--> ├──────┼──────────┼───────────┼────────┼─────────────────┤
--> │      │ 1,010    │ 0         │ false  │ sleep 1 && exit │
--> │      │ 5        │ 0         │ false  │ exit            │
--> └──────┴──────────┴───────────┴────────┴─────────────────┘
[1] exit started at 2022-04-24 19:28:03.249
[1] exit stopped at 2022-04-24 19:28:03.254 after 5ms
[1] exit exited with code 0
```

kayahr:
This fixes the problem that summary timings are printed before grouped output (Fixes #316).

### DIFF
--- a/src/flow-control/log-timings.spec.ts
+++ b/src/flow-control/log-timings.spec.ts
@@ -57,12 +57,14 @@ it('returns same commands', () => {
 
 it("does not log timings and doesn't throw if no logger is provided", () => {
     controller = new LogTimings({});
-    controller.handle(commands);
+    const { onFinish } = controller.handle(commands);
 
     commands[0].timer.next({ startDate: startDate0 });
     commands[1].timer.next({ startDate: startDate1 });
     commands[1].timer.next({ startDate: startDate1, endDate: endDate1 });
     commands[0].timer.next({ startDate: startDate0, endDate: endDate0 });
+
+    onFinish?.();
 
     expect(logger.logCommandEvent).toHaveBeenCalledTimes(0);
 });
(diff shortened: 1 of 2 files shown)


kayahr:
Hm... No idea why CI test failed on Windows with Node.js 14 but succeeded on all other platforms including Windows with Node.js 16. Maybe this is a CI environment problem? Huge load on the CI node instance so it ran into a timeout? Can the test be repeated somehow?

paescuj:
> Hm... No idea why CI test failed on Windows with Node.js 14 but succeeded on all other platforms including Windows with Node.js 16. Maybe this is a CI environment problem? Huge load on the CI node instance so it ran into a timeout? Can the test be repeated somehow?

Yeah, this is a timeout we unfortunately run into from time to time... Restarted the failing test 👍

paescuj (approved):
Thanks for your contribution ❤️ LGTM!

Issue #437, opened 2023-09-02
No colors for prefixes when using `prefixColors: "auto"` via concurrent API


OnkelTem:
concurrently: 8.2.1
node: v16.17.0

I use the following script to run concurrent processes:

```js
const fs = require('fs')
const path = require('path')
const concurrently = require('concurrently')

const workspacePaths = JSON.parse(fs.readFileSync('./package.json')).workspaces

const { result } = concurrently(
  workspacePaths.map((wp) => ({
    command: 'npm:dev',
    name: path.basename(wp),
    cwd: path.resolve(__dirname, '../', wp),
  })),
  {
    prefix: 'name',
    killOthers: ['failure', 'success'],
    restartTries: 0,
    prefixColors: 'auto',
  }
)

result.then(
  () => {
    console.log('Success')
  },
  () => {
    console.log('Failed')
  }
)
```

However, when I run in the terminal, it doesn't show any colors in the prefixes:

![image](https://github.com/open-cli-tools/concurrently/assets/114060/debfed7f-74ae-4307-8590-c19166c27601)

yet other colors are displayed. What could be the problem?

Thanks in advance.

SeiwonPark:
Check out API's `prefixColors`. It says

### Docs
<p><code>prefixColors: <b>a list of colors</b> as supported by chalk or auto for an automatically picked color. </code></p>

### Definition
```
prefixColors?: string[]
```

<br />   

So change it to 

```js
prefixColors: ['auto'],
```

paescuj:
I think @SeiwonPark is right! (Thanks!)
Therefore I'm closing this issue for now, but let me know if that doesn't work either and we'll reopen it.

OnkelTem:
Thanks @SeiwonPark

However, an update to the docs was made:

> `prefixColors`: a list of colors or a string as supported by [chalk](https://www.npmjs.com/package/chalk) and additional style `auto` for an automatically picked color.

And I support it. Now it sounds much better.

PR #439, opened 2023-09-07
fix(suggestion): prefixColors API type to embrace string type

SeiwonPark:
## Purpose of this PR
Regarding [`prefixColors`'s default value](https://github.com/open-cli-tools/concurrently/blob/16182089336827900ef2fc943b353c14b850cc9b/src/defaults.ts#L38), and the issue #437, though the document specified as below,

```
a list of colors as supported by chalk or auto for an automatically picked color.
```

still the behavior for using API doesn't look quite neat. Here's an example:

### example
```js
// referred code from README

const { result } = concurrently(
  [
    'npm:watch-*',
    { command: 'nodemon', name: 'server' },
    { command: 'deploy', name: 'deploy', env: { PUBLIC_KEY: '...' } },
    {
      command: 'watch',
      name: 'watch',
      cwd: path.resolve(__dirname, 'scripts/watchers'),
    },
  ],
  {
    prefix: 'name',
    killOthers: ['failure', 'success'],
    restartTries: 3,
    cwd: path.resolve(__dirname, 'scripts'),
    prefixColors: ['auto'],     //  <--------- To use 'auto' option, it needs to be wrapped in a list
  },
);
result.then(success, failure);
```

and as the option is `prefixColor"s"`, it makes sense. 

But it's literally an `auto` so rather than wrapping it in a list, 

```js
prefixColors: 'auto' // or 'bgMagenta', 'red', 'underline', ... for other available styles
```

just give it a string like above might not hurt the philosophy of [`prefixColors`'s default value](https://github.com/open-cli-tools/concurrently/blob/16182089336827900ef2fc943b353c14b850cc9b/src/defaults.ts#L38).

So it might be another handy option for users I guess. Or how do you think about this?

<br />   

## What's been updated
So I'm suggesting the `string` type to be able to embrace all the behaviors from users.
```diff
// src/concurrently.ts

- prefixColors?: string[];
+ prefixColors?: string | string[];
```

<br />   

concurrently

and updated document for more details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,7 +335,7 @@ For more details, visit https://github.com/open-cli-tools/concurrently
   - `prefix`: the prefix type to use when logging processes output.
     Possible values: `index`, `pid`, `time`, `command`, `name`, `none`, or a template (eg `[{time} process: {pid}]`).
     Default: the name of the process, or its index if no name is set.
-  - `prefixColors`: a list of colors as supported by [chalk](https://www.npmjs.com/package/chalk) or `auto` for an automatically picked color.
+  - `prefixColors`: a list of colors or a string as supported by [chalk](https://www.npmjs.com/package/chalk) and additional style `auto` for an automatically picked color.
     If concurrently would run more commands than there are colors, the last color is repeated, unless if the last color value is `auto` which means following colors are automatically picked to vary.
     Prefix colors specified per-command take precedence over this list.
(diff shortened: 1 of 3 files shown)


gustavohenke (approved):
Thanks, LGTM.

paescuj (approved):
Thanks!

Issue #399, opened 2022-12-31
Import footgun: Default export versus named export (No output when used via API)


rezonant:
I got bit by accidentally importing as `import { concurrently } from 'concurrently';` instead of `import concurrently from 'concurrently';`. The default export applies all the default controllers and logger, whereas the named export does not (the default export uses the "unconfigured" named export to do it's thing). 

It might be worth adding a note to the README about this or possibly renaming the named export to make it harder to fall into this trap. 

The symptoms are that no program output appears, even though it does when you use the concurrently CLI- this is of course because there is no Logger or LogOutput/LogError controllers added in the configuration.
I ended up having to dig into the source code to figure this out.

At the very least, this issue should provide some help for anyone else who hits this footgun.

gustavohenke:
You've raised an excellent point!
I suppose that renaming the unconfigured function to `createConcurrently` would be better, but that'd be a breaking change 😕

paescuj:
> I suppose that renaming the unconfigured function to `createConcurrently` would be better, but that'd be a breaking change 😕

Might be something for v8?

gustavohenke:
Hey! This is now fixed in v9.0.0.
https://github.com/open-cli-tools/concurrently/releases/tag/v9.0.0

PR #440, opened 2023-09-10
Fix default export returning undefined in ESM 

Codex-:
I noticed this previously but it's consistently reproducible with Bun.

The esm based `index.mjs` export for `concurrently`, `export default concurrently.default`, in some cases depending on the way you're executing `concurrently` can have the default function on `concurrently` itself which makes this export not work as expect, as `default` is undefined.

This PR adds a simple fallback `export default concurrently.default || concurrently` that allows _some_ defense against this behaviour.

I have added tests for both Node and Bun that test the `commonjs` and `esm` import behaviour, as well as a direct `ts` import case testing both for Bun's case.

Also, I've noticed Node 19 is targetted here but Node 20 enters LTS next month, should this target 20 instead of 19?

### DIFF
--- a/smoke-tests/smoke-test-cjs.cjs
+++ b/smoke-tests/smoke-test-cjs.cjs
@@ -4,21 +4,11 @@
 const assert = require('node:assert');
 const concurrently = require('../index.js');
 
-(async () => {
-  try {
-    // Assert the functions loaded by checking their names load and types are correct
-    assert.strictEqual(
-      typeof concurrently === 'function',
-      true,
-      'Expected default to be function',
-    );
+// Assert the functions loaded by checking their names load and types are correct
+assert.strictEqual(
+  typeof concurrently === 'function',
+  true,
+  'Expected default to be function',
+);
 
-    console.info('Imported cjs successfully');
-  } catch (error) {
-    console.error(error);
-    console.debug(error.stack);
-
-    // Prevent an unhandled rejection, exit gracefully.
-    process.exit(1);
-  }
-})();
+console.info('Imported cjs successfully');
--- a/smoke-tests/smoke-test-cjs.cjs
+++ b/smoke-tests/smoke-test-cjs.cjs
@@ -1,0 +1,24 @@
+// @ts-check
+/* eslint-disable @typescript-eslint/no-var-requires, no-console */
+
+const assert = require('node:assert');
+const concurrently = require('../index.js');
+
+(async () => {
+  try {
+    // Assert the functions loaded by checking their names load and types are correct
+    assert.strictEqual(
+      typeof concurrently === 'function',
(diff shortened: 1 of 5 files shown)


smoke-tests/smoke-test-cjs.cjs
```diff
@@ -0,0 +1,24 @@
...
+      'Expected default to be function',
+    );
+
+    console.info('Imported cjs successfully');
+  } catch (error) {
+    console.error(error);
+    console.debug(error.stack);
+
+    // Prevent an unhandled rejection, exit gracefully.
+    process.exit(1);
```
gustavohenke, line 22: Is there a difference?

smoke-tests/smoke-test-cjs.cjs
```diff
@@ -0,0 +1,24 @@
...
+    assert.strictEqual(
+      typeof concurrently === 'function',
+      true,
+      'Expected default to be function',
+    );
+
+    console.info('Imported cjs successfully');
+  } catch (error) {
+    console.error(error);
+    console.debug(error.stack);
```
gustavohenke, line 19: Node logs the stacktrace by default, doesn't bun do the same? E.g. if you make the assertion fail, won't it print the stack trace twice?

smoke-tests/smoke-test-cjs.cjs
```diff
@@ -0,0 +1,24 @@
...
+  try {
+    // Assert the functions loaded by checking their names load and types are correct
+    assert.strictEqual(
+      typeof concurrently === 'function',
+      true,
+      'Expected default to be function',
+    );
+
+    console.info('Imported cjs successfully');
+  } catch (error) {
```
gustavohenke, line 17: Why is catching the error necessary?

gustavohenke (commented):
Thanks for fixing this! I have mostly nitpicks, no questions on the intended change

smoke-tests/smoke-test-cjs.cjs
Codex-: non-zero exit codes will fail the CICD, there were cases in my own testing where simply using `process.exit()` would cause a false success on the workflow run

smoke-tests/smoke-test-cjs.cjs
Codex-: unhandled rejections can fail silently, but these don't actually need to be async so I'll update them (ported from a larger suite I use for one of my own packages)

paescuj:
Not sure if this is the preferred solution. I think this could be properly tackled by https://github.com/open-cli-tools/concurrently/issues/399 instead.
Also I think it's about time to go full ESM, things are getting messy.
I'll look into this more detailed later on this day.

Issue #444, opened 2023-09-27
CLI options not clear in README


Tarektouati:
Hi ! 
I recently setup  concurrently to speed up my some command execution on my CI.
I did find the `Help` section quite useful, but confusing about the possible usage.  
This is more of a suggestion, maybe we can rewrite that section. 

Thanks

paescuj:
Suggestions and pull requests are most welcome!
IMO the whole README deserves a revision 😃 Would also nice to automate the generation of the API section from types.

gustavohenke:
The CLI documentation has improved a lot lately, so I'm closing this issue for now.
Will look into merging the linked PR separately.

PR #445, opened 2023-10-01
Modify cli options and API in README

Tarektouati:
This PR closes issue #444

It aims to 
- Simplify usage and CLI options in the README.
- Automate the generation of API section

### TODO :
- [x]  Modify "CLI options" section  
- [x]  Generate "API" section
- [ ]  Automate API section automation


README.md
```diff
@@ -138,282 +141,386 @@ concurrently "npm:lint:*(!fix)"
...
+
+A command parser encapsulates a specific logic for mapping `CommandInfo` objects
+into another `CommandInfo`.
+
+A prime example is turning an abstract `npm:foo` into `npm run foo`, but it could also turn
+the prefix color of a command brighter, or maybe even prefixing each command with `time(1)`.
+
+**Members:**
+
+- /\*\*
```
SeiwonPark, line 366: [ts-readme](https://github.com/intuit/ts-readme) doesn't seem to be maintained and it doesn't seem to parse inner jsdoc like [this](https://github.com/open-cli-tools/concurrently/blob/main/src/command-parser/command-parser.ts#L11)

README.md
```diff
@@ -432,3 +539,11 @@ It contains the following properties:
 - Does this work with the npm-replacements [yarn](https://github.com/yarnpkg/yarn), [pnpm](https://pnpm.js.org/), or [Bun](https://bun.sh/)?
 
   Yes! In all examples above, you may replace "`npm`" with "`yarn`", "`pnpm`", or "`bun`".
+
+```
```
SeiwonPark, line 543: this and some others (Commented since `Generate "API" section` has been marked)

README.md
SeiwonPark: How about using [typedoc](https://www.npmjs.com/package/typedoc-plugin-markdown)?

Tarektouati: I didn't knew this package,
It seems pretty nice and match our use-case here.

I'll update the PR with it.

README.md
Tarektouati: It's caused by `ts-readme`, it will change as we'll use an other package.

src/index.ts
```diff
@@ -18,11 +18,9 @@ import { LogTimings } from './flow-control/log-timings';
 import { RestartProcess } from './flow-control/restart-process';
 import { Logger } from './logger';
 
+/** Logger options  */
```
gustavohenke, line 21: I didn't go through the .md file yet, but it's my perception (and, well, it's what we do at Canva too) that `//` comments are usually not picked up by tooling (doc generators and IDEs), whereas `/* */` comments are, so they work mostly as a file-internal comment that doesn't need to be exposed.

The old `// Logger options` was here mostly to create a section (just like there are others for input handling, restarting and so on below), I don't think it adds any value being in `/* */` nor is it correct as the doc for `ConcurrentlyOptions`

cc @paescuj just because I think this has been an unwritten rule in this codebase :)

gustavohenke (commented):
I like very much the new tables for each of the options! I think it might be possible to autogenerate those too (by inspecting yargs state).

I'm not sure I like how the TS docs generation is structured in the readme; it might be too much now. Maybe the typedoc suggestion (https://github.com/open-cli-tools/concurrently/pull/445#discussion_r1343516868) is good, but I haven't spent much more time looking into it. Will do once the PR is ready for review.

README.md
Tarektouati: @SeiwonPark I've been quite busy lately.

I've implemented `typedoc` let me know what do you think before I can start working on automation.

src/index.ts
Tarektouati: I modified theses comments for `ts-readme` 

they should be reverted after migrating to `typedoc` : (cf: https://github.com/open-cli-tools/concurrently/pull/445#discussion_r1343516868)

docs/.nojekyll
```diff
@@ -0,0 +1 @@
+TypeDoc added this file to prevent GitHub Pages from using Jekyll. You can turn off this behavior by setting the `githubPages` option to false.
```
Tarektouati, line 1: This file is generated by `typedoc`

docs/classes/Command.md
```diff
@@ -0,0 +1,264 @@
+# Class: Command
```
Tarektouati, line 1: Moved to separate folder `docs`

Updated links in the `README.md` to this file as entry point for documentation

scripts/generate-documentation.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+#Generate documentation for the project based on Typescript types
+pnpm exec typedoc
+
+# Typedoc generates a README.md file in the docs folder, which is not needed
```
Tarektouati, line 5: I think the comment is explicit here,
I couldn't find a way to avoid `typedoc` to regenerate `README.md` file in `docs` folder

src/index.ts
Tarektouati: Reverted in 982c15fcfb1e747742f0e45a5fbf14b1028ccdac

README.md
SeiwonPark: cc @gustavohenke @paescuj 

Pretty nice job! [Interfaces](https://github.com/Tarektouati/concurrently/blob/main/docs/interfaces/CloseEvent.md) are documented pretty nicely but not sure if [classes](https://github.com/Tarektouati/concurrently/blob/main/docs/classes/Command.md) are. (And some other fixes are required as well)

IMHO as @gustavohenke liked tables, current table-sorted [README](https://github.com/Tarektouati/concurrently/blob/main/README.md) style seems better. And there seems to be more configuration to do to automate the process. So how about just let maintainers/contributors manually update the document more readable in current stage?

I've recommended **typedoc** to be like some use cases (_e.g. [chatgpt-api](https://github.com/transitive-bullshit/chatgpt-api/blob/main/docs/readme.md), and [mermaid](https://github.com/mermaid-js/mermaid/blob/develop/docs/intro/index.md), ... and some other open sources_). **typedoc** could be an option but this might be too early for `concurrently` in this stage(because to use it like the above libraries, pretty much comments should be fixed and written to fit the **typedoc** configuration).

So maybe setting the issue(**automating documentation generation**) as `<Pending>` and keep @Tarektouati's current README style that contains table-sorted document (but yes update document manually..) would be enough. Or, rather implementing separate library _e.g. some-library-like-jsdoc-to-readme-parser_ would be a nice choice.

Tarektouati: Thanks ! 

Totally agree with @SeiwonPark, as I see the generation result, I'm not totally convinced. 
At least, it shows where codebase should be enhanced to have a better experience with it. 

Let me know if we keep "automating documentation generation" task and create other PR in that direction, or maybe just drop the related commits.

cc @gustavohenke @paescuj

Issue #433, opened 2023-07-09
`--kill-others-on-fail` flag not working as expected with `--group --max-processes 1`


daniil4udo:
### Description

When trying to run multiple scripts sequentially using concurrently, I noticed that the `--kill-others-on-fail` flag is not behaving as expected when used together with the `--group --max-processes 1` flags, rest of the scripts still running.

### Expected Behavior

If one script fails (exits with a non-zero code), all running processes should be terminated when using the `--kill-others-on-fail` flag, even when running scripts sequentially with `--group --max-processes 1`.

### Current Behavior

However, when I use the `--kill-others-on-fail` flag in conjunction with `--group --max-processes 1`, concurrently is not terminating all running processes when one script fails. This behaviour is not observed when I remove --group --max-processes 1 from the command.

gustavohenke:
Thanks for the report! I think this use case makes sense.

There's a potentially funny interaction with `--restart-tries` here. The first command will restart N times before giving up, but should other commands be given a chance to spawn at all?

bozdoz:
How about a `--kill-others-on-sigint`?  I'm not sure I understand the interaction with "restart" here.  If the user cancels the process, I'd expect the main process (concurrently) to stop the subprocesses.

Don't the restart flags default to `0`?

gustavohenke:
> Don't the restart flags default to 0?

Yes.

The issue I see starts like this:

```
concurrently --kill-others-on-fail --max-processes 1 "sleep 5 && echo foo" "exit 1"
```

- **Current behaviour** is that command 1 exits, and a SIGTERM (from `--kill-others-on-fail`) is sent to nowhere, so command 0 still runs
- **New behaviour** would be that command 0 shouldn't be given a chance to run, since it was meant to be killed anyways.

But once you add `--restart-tries 1`, then it's fair that command 0 gets a chance to run... or not really?

bozdoz:
Should restart give the commands a chance to run? No. Not on SIGINT. The user is aborting the process. 

Thinking of docker restart policies:
- on-failure
- always
- unless-stopped

None of these restart when the container is manually stopped: even "always" explicitly doesn't include if it's manually stopped.

gustavohenke:
Yeah, SIGINT has a quite clear use case.

[...]

PR #455, opened 2023-12-12
Add a state property to Command

gustavohenke:
Title.
The intention is that this will make it easier to tell exactly where the command is. My idea is that it can be used for #433, since it might be necessary to make a distinction between a command that was never started, and everything else.

Also marked it a breaking change because I removed the `#exited` property, as `#state` supersedes it.
Booleans do a poor job sometimes :)

Most LOC in this PR are coming from moving a few tests around.


gustavohenke:
happy new year 🥳

PR #456, opened 2023-12-14
Export named concurrently + fix TS errors with Node16

### LINKED ISSUE #399: Import footgun: Default export versus named export (No output when used via API)

### LINKED ISSUE #454: Error TS2349 using concurrently in full ESM module
Hello, 

I have a problem using  concurrently (and others modules) in my project after migrate from CJS to full ESM. I have the error TS2349 when I try to call `concurrently` function. But, In VSCode, the definition is correctly loaded on CTRL+click :/

This is my config : 

- tsconfig.json
    ```json
    {
      "compileOnSave": false,
      "compilerOptions": {
        "rootDir": ".",
        "sourceMap": true,
        "declaration": true,
        "esModuleInterop": true,
        "emitDecoratorMetadata": true,
        "experimentalDecorators": true,
        "importHelpers": false,
        "resolveJsonModule": true,
        "noImplicitAny": false,
        "strict": true,
        "useDefineForClassFields": false,
        "target": "ES2022",
        "lib": [
          "esnext"
        ],
        "skipLibCheck": true,
        "skipDefaultLibCheck": true,
        "baseUrl": ".",
        "outDir": "dist",
        "types": [
          "node"
        ],
        "module": "NodeNext",
        "moduleResolution": "NodeNext",
      }
    }
    ```

And my example code : 

```ts
import concurrently from 'concurrently';

const main = async () => {
    concurrently([
        { command : 'ls /etc' },
        { command : 'ls /usr' }
    ]);
}

main();
```

He is a screen on my VSCode : 
image

I don't know how to do to fix that. I think my problem is with typescript, but I don't understand how to fix that :/

gustavohenke:
Fixes a bunch of issues when using concurrently
1. For `require()` function (from both TS and pure JS)
2. With TS + import syntax, but without ES module interop
3. With TS + import syntax, but with ES module interop
4. As an ES Module (from both TS and pure JS)

On the side, the named export `concurrently` is now the fully configured concurrently.
This was done to simplify the exports in the index files (most likely specific to ESM, but I've lost track of why exactly).

Will add some smoke tests in a follow up PR so that this one stays mostly focused to the core changes.

Fixes #399
Fixes #454
Closes #440

### DIFF
--- a/bin/concurrently.ts
+++ b/bin/concurrently.ts
@@ -3,7 +3,7 @@ import yargs from 'yargs';
 import { hideBin } from 'yargs/helpers';
 
 import * as defaults from '../src/defaults';
-import concurrently from '../src/index';
+import { concurrently } from '../src/index';
 import { epilogue } from './epilogue';
 
 // Clean-up arguments (yargs expects only the arguments after the program name)
--- a/index.d.mts
+++ b/index.d.mts
@@ -1,0 +1,7 @@
+/*
+ * While in local development, make sure you've run `pnpm run build` first.
+ */
+import { concurrently } from './dist/src/index.js';
+
+export * from './dist/src/index.js';
+export default concurrently;
--- a/index.d.mts
+++ b/index.d.mts
@@ -1,7 +1,4 @@
-/*
- * While in local development, make sure you've run `pnpm run build` first.
- */
-import { concurrently } from './dist/src/index.js';
+import { concurrently } from './src/index.js';
 
-export * from './dist/src/index.js';
+export * from './src/index.js';
 export default concurrently;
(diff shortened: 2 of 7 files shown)


gustavohenke:
Bit of a bump @paescuj, might proceed without a review in a few days

gustavohenke:
Hey! This is now fixed in v9.0.0.

PR #460, opened 2024-01-10
Abort commands not running when max processes < N

### LINKED ISSUE #433: `--kill-others-on-fail` flag not working as expected with `--group --max-processes 1`

### LINKED ISSUE #452: Cancel a series of tasks with SIGINT
Description:

I want to run a series of tasks, that are cancellable.

Expected Behaviour:

If I press CTRL-C, I want the entire script (each of the series of tasks) to be aborted (or not started).

Actual Behaviour:

I need to press CTRL-C multiple times to cancel each process.

Reproduction:

Note the `-m 1` flag to run one process consecutively.

```json
{
"sleep": "concurrently --kill-others-on-fail -m 1 \"sleep 3\" \"echo 'hi'\" \"sleep 3\" \"echo 'there'\" \"sleep 3\" \"echo 'buddy'\""
}
```

```bash
npm run sleep

> concurrently --kill-others-on-fail -m 1 "sleep 3" "echo 'hi'" "sleep 3" "echo 'there'" "sleep 3" "echo 'buddy'"

^C[0] sleep 3 exited with code SIGINT
[1] hi
[1] echo 'hi' exited with code 0
[2] sleep 3 exited with code 0
[3] there
[3] echo 'there' exited with code 0
[4] sleep 3 exited with code 0
[5] buddy
[5] echo 'buddy' exited with code 0
```

Is this a bug, or is there a way to do this that is not obvious?

gustavohenke:
## Problem
In a few issues, this situation came up where commands that haven't started yet will run anyways when they shouldn't even start.
Best examples of this are SIGINT and `--kill-others` combined with `--max-processes` that's lower than N (where N = number of commands itself).

For someone who pressed <kbd>Ctrl</kbd> + <kbd>C</kbd>, it could be annoying that one press isn't sufficient and they have to press it up to N times (see #452).

For `--kill-others`/`--kill-others-on-fail`, it, well... just doesn't work (see #433).

## Solution

With SIGINT, it's quite clear that commands awaiting their turn should be aborted.

For `--kill-others`, there was a possible interaction with `--restart-tries`: should it be respected at all?
But the flag is documented as _"restart processes that died"_, which isn't the case of a command that never started. Therefore I made the decision of also aborting these commands.

Now, aborting the commands is done here by using an `AbortController`.
It's quite simple: flow controllers make use of it, and concurrently stops further spawning of processes if the signal is already aborted.

> [!NOTE]
> `AbortController` is built into Node 15+, which is a requirement for concurrently 9.0.0 (#448), therefore this PR is a breaking change.

---

Fixes #433 
Fixes #452

### DIFF
--- a/src/completion-listener.spec.ts
+++ b/src/completion-listener.spec.ts
@@ -47,7 +47,8 @@ describe('listen', () => {
 
     it('completes when abort signal is received and command is stopped, returns nothing', async () => {
         const abortCtrl = new AbortController();
-        const result = createController('all').listen([new FakeCommand()], abortCtrl.signal);
+        // Use success condition = first to test index access when there are no close events
+        const result = createController('first').listen([new FakeCommand()], abortCtrl.signal);
 
         abortCtrl.abort();
         scheduler.flush();
--- a/src/completion-listener.spec.ts
+++ b/src/completion-listener.spec.ts
@@ -24,6 +24,31 @@ const createController = (successCondition?: SuccessCondition) =>
 const emitFakeCloseEvent = (command: FakeCommand, event?: Partial<CloseEvent>) =>
     command.close.next(createFakeCloseEvent({ ...event, command, index: command.index }));
 
+it('completes only when commands emit a close event, returns close event', async () => {
+    const abortCtrl = new AbortController();
+    const result = createController('all').listen(commands, abortCtrl.signal);
+
+    commands[0].state = 'started';
+    abortCtrl.abort();
+
+    const event = createFakeCloseEvent({ exitCode: 0 });
+    commands[0].close.next(event);
+    scheduler.flush();
+
+    await expect(result).resolves.toHaveLength(1);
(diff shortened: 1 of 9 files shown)


gustavohenke:
@paescuj keen to take a look?

gustavohenke:
🚢 This is now fixed in v9.0.0!

Issue #463, opened 2024-01-31
Programmatically restarting Command using `kill/start` API produce wrong termination


odeadglaz:
**Description**
I'm using `concurrently` to run 2 commands in parallel on my application:
1. Watch's application code using `swc`
2. A plain `node ./dist/server.js` command which starts my HTTP server.

I wanted to implement a logic of which every time `SWC` outputs a successful watch compilation output, to restart the `NodeJS` server by combining the `Command` object `kill/start` API.

Here is a simplified example of my code:
```ts
const commands = [
    { name: 'SWC watcher', command: 'swc src -d dist/server -w -D' },
    { name: 'HTTP Server', command: 'node -r dotenv/config ./dist/server/index.js' },
];

const concurrency = concurrently(commands, {
    killOthers: ['failure'],
    prefixColors: ['magenta', 'yellow']
});

let shouldRestart = false;
const [watcherCommand, serverCommand] = concurrency.commands;

serverCommand.close.subscribe(() => {
    if (shouldRestart) {
        shouldRestart = false;
        serverCommand.start();
    }
});

watcherCommand.stdout.subscribe((data) => {
    const input = data.toString().trim();

    if (compilationCompleteRegex.test(input)) { // Detects compilation end output
        shouldRestart = true;
        serverCommand.kill('SIGTERM'); // HTTP listens and gracefully shuts down
    }
});
```

**Actual Behaviour**
After 2 restarts the `concurrency` top command **resolves** as both commands was terminated but in fact, both commands still up and running.

**Some Technical details**
It seems that `CompletionListener.listen` method actually uses [`bufferCount`](https://github.com/open-cli-tools/concurrently/blob/main/src/completion-listener.ts#L92) to assume its commands were terminated by counting the `close` events but programmatically using `Command.start`/`Command.kill` API's can cause un-expected behaviour in this manner as nothing was actually terminated rather solely **restarted**

Thanks in advanced 🙏

odeadglaz:
Hey, I opened a draft PR for a potential fix [here](https://github.com/open-cli-tools/concurrently/pull/464) would love for any comments 🙏

gustavohenke:
I recommend using something like [node-dev](https://www.npmjs.com/package/node-dev).

If you don't want that, or it doesn't work for some reason, another solution you could use today is to set `restartTries: -1`. This makes killed processes restart forever, so you don't need to manually restart. Completion is autmoatically handled in this case.

[...]

PR #464, opened 2024-02-01
Ensuring completion-listener.ts listen to a single close event emitte…

odeadglaz:
Currently, the `CompletionListener` sets a `bufferCount` upon commands close streams to conclude once commands are finished and then evaluates the exit code and determines the `success/fail` conditions.

Continued to the issue brought [here](https://github.com/open-cli-tools/concurrently/issues/463), calling manually `Command` - `kill/start` API's manually, causes a bug where the `CompletionListener` resolves before all commands had exited ( as a single command can emit multiple close events ).

This PR changes the behaviour to take a single emit event from each command in order to conclude termination.

**Disclaimer:**
Potentially I would love to find a way to take the **last** exit command emitted from each `Command` tho I couldn't find a way would appreciate any suggestion 👍 mine while I kept things the same.

**Side effects**
- I re-arranged the `completion-listener.spec.ts` file to all use the same command exit helper
- ~~Allow `ipc` for spawned commands~~

Resolves: https://github.com/open-cli-tools/concurrently/issues/463

### DIFF
--- a/src/completion-listener.spec.ts
+++ b/src/completion-listener.spec.ts
@@ -24,6 +24,8 @@ const createController = (successCondition?: SuccessCondition) =>
 const emitFakeCloseEvent = (command: FakeCommand, event?: Partial<CloseEvent>) =>
     command.close.next(createFakeCloseEvent({ ...event, command, index: command.index }));
 
+const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+
 describe('listen', () => {
     it('check for success once all commands have emitted at least a single close event', async () => {
         const finallyCallback = jest.fn();
@@ -36,7 +38,7 @@ describe('listen', () => {
 
         scheduler.flush();
 
-        await Promise.resolve();
+        await sleep(100);
         expect(finallyCallback).not.toHaveBeenCalled();
 
         commands[1].close.next(createFakeCloseEvent({ exitCode: 0 }));
@@ -47,6 +49,19 @@ describe('listen', () => {
         await expect(result).resolves.toEqual(expect.anything());
 
         expect(finallyCallback).toHaveBeenCalled();
+    });
+
+    it('Takes last event emitted from each command', async () => {
+        const result = createController().listen(commands);
+
+        commands[0].close.next(createFakeCloseEvent({ exitCode: 0 }));
+        commands[0].close.next(createFakeCloseEvent({ exitCode: 1 }));
+        commands[1].close.next(createFakeCloseEvent({ exitCode: 0 }));
+        commands[2].close.next(createFakeCloseEvent({ exitCode: 0 }));
+
+        scheduler.flush();
+
+        await expect(result).rejects.toEqual(expect.anything());
     });
 });
 
@@ -173,7 +188,6 @@ describe('Detect commands exit conditions', () => {
         });
 
         it(`fails if some commands ${nameOrIndex} exit with non-0 code`, () => {
-            commands = [commands[0], commands[1], commands[1]];
             const result = createController(condition).listen(commands);
 
             emitFakeCloseEvent(commands[0], { exitCode: 1 });
--- a/src/completion-listener.spec.ts
+++ b/src/completion-listener.spec.ts
@@ -24,7 +24,7 @@ const createController = (successCondition?: SuccessCondition) =>
 const emitFakeCloseEvent = (command: FakeCommand, event?: Partial<CloseEvent>) =>
     command.close.next(createFakeCloseEvent({ ...event, command, index: command.index }));
 
-const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+const flushPromises = () => new Promise((resolve) => setTimeout(resolve, 0));
 
 describe('listen', () => {
     it('check for success once all commands have emitted at least a single close event', async () => {
@@ -37,8 +37,8 @@ describe('listen', () => {
         commands[0].close.next(createFakeCloseEvent({ exitCode: 0 }));
 
         scheduler.flush();
-
-        await sleep(100);
+        // A broken implementantion will have called finallyCallback only after flushing promises
+        await flushPromises();
         expect(finallyCallback).not.toHaveBeenCalled();
 
         commands[1].close.next(createFakeCloseEvent({ exitCode: 0 }));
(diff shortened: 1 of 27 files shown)


odeadglaz:
Rethinking on this, this change includes also an edge case that doesn't work:

- Starting 2 commands:
- First command manually killed, and then started
- The second command manually killed

---> It will resolve altho the first command is also alive

So I might be going on the wrong direction here would love to here your tought on the initial design of the `start`/`kill` API, were they meant to be used this way?

odeadglaz:
Another option would be a design-level change such as:

Adding a new `restart` API for command which goes something like:
```
restart(code? :string) {
    this.state = 'restart'; // New CommandState
    this.kill();
    this.start();
} 
````

And a modification for `Command.start` for skipping emitting close event if the command state equals to `restart`

Would love to hear your mind on it 👍

**Review by gustavohenke — COMMENTED** [review]
Restarting is tricky; because each option you set changes the closing conditions of a command, the flow controllers wrap the close observable so that all looks simple on the completion listener side.

> Adding a new restart API

I think this could be done for convenience, but not as a way of fixing this issue.

> this change includes also an edge case that doesn't work:

This is true, but this PR is at least an improvement over not doing anything.

The optimal solution is probably to retain the last emitted close event of each command until they all exit.

**Comment by odeadglaz** [comment]
So how would u recommend proceeding? Do you wish to have this fix for now? (as you said just a slight improvement for the current state)

File: src/completion-listener.ts [hunk 1]
```diff
@@ -86,7 +86,8 @@ export class CompletionListener {
      * @returns A Promise that resolves if the success condition is met, or rejects otherwise.
      */
     listen(commands: Command[]): Promise<CloseEvent[]> {
-        const closeStreams = commands.map((command) => command.close);
+        const closeStreams = commands.map((command) => command.close.pipe(take(1)));
+
         return Rx.lastValueFrom(
             Rx.merge(...closeStreams).pipe(
                 bufferCount(closeStreams.length),
```
gustavohenke, line 93: > So how would u recommend proceeding?

I think that replacing these lines with `Rx.combineLatest()` + removing the `take(1)` above can achieve retaining the last emitted close event.

We can then also fix your edge case from https://github.com/open-cli-tools/concurrently/pull/464#issuecomment-1921230207 with a `filter()` operator that checks if every command is indeed closed.
This means waiting until every command emits another close event.

WDYT?

odeadglaz: Yea that sounds legit, at-least the first part, seems like `combineLatest` can do the job 👍  Ill try to change the implementation accordingly

odeadglaz: @gustavohenke I've pushed the changes, the only drawback using the `combineLatest` is that it actually keep the close events according to the registered stream **and not by the actual order they were emitted**, So beside accepting your offer I also added a `map` to stort it via the `startingDate` of the event, would love to hear if you got any better suggestion here 👍

src/completion-listener.ts
```diff
@@ -86,10 +86,18 @@ export class CompletionListener {
...
      */
     listen(commands: Command[]): Promise<CloseEvent[]> {
-        const closeStreams = commands.map((command) => command.close);
+        const closeStreams = commands.map((command) => command.close.pipe());
+
         return Rx.lastValueFrom(
-            Rx.merge(...closeStreams).pipe(
-                bufferCount(closeStreams.length),
+            Rx.combineLatest(closeStreams).pipe(
+                filter((exitInfos) => exitInfos.every((exitInfo) => exitInfo.exitCode !== null)),
```
gustavohenke, line 93: `exitCode` is never null here. You need to look at the event's command for the current state

```suggestion
                filter((exitInfos) => exitInfos.every(({ command }) => command.state !== 'started')),
```

src/completion-listener.ts
```diff
@@ -86,10 +86,18 @@ export class CompletionListener {
      * @returns A Promise that resolves if the success condition is met, or rejects otherwise.
      */
     listen(commands: Command[]): Promise<CloseEvent[]> {
-        const closeStreams = commands.map((command) => command.close);
+        const closeStreams = commands.map((command) => command.close.pipe());
```
gustavohenke, line 89:
```suggestion
        const closeStreams = commands.map((command) => command.close);
```

src/completion-listener.spec.ts
```diff
@@ -24,186 +24,228 @@ const createController = (successCondition?: SuccessCondition) =>
...
         scheduler.flush();
 
-        return expect(result).resolves.toEqual(expect.anything());
+        await expect(result).resolves.toEqual(expect.anything());
+
+        expect(finallyCallback).toHaveBeenCalled();
     });
 
-    it('fails if one of the processes exited with non-0 code', () => {
+    it('Takes last event emitted from each command', async () => {
```
gustavohenke, line 54: nitpicking 🤓 

```suggestion
    it('takes last event emitted from each command', async () => {
```

src/completion-listener.ts
```diff
@@ -86,10 +86,18 @@ export class CompletionListener {
...
+
         return Rx.lastValueFrom(
-            Rx.merge(...closeStreams).pipe(
-                bufferCount(closeStreams.length),
+            Rx.combineLatest(closeStreams).pipe(
+                filter((exitInfos) => exitInfos.every((exitInfo) => exitInfo.exitCode !== null)),
+                map((exitInfos) =>
+                    exitInfos.sort((first, second) => {
+                        return (
+                            first.timings.startDate.getTime() - second.timings.startDate.getTime()
```
gustavohenke, line 97: The tests are passing, but they shouldn't.

The definition of conditions `first` and `last` is "the first/last command to exit", but in tests the events are emitted in the tested order, and their start date is the current date. Hence this line didn't break any tests.

In reality though, command 1 might start earlier, but exit later than command 2, so this needs to be

```suggestion
                            first.timings.endDate.getTime() - second.timings.endDate.getTime()
```

gustavohenke (changes requested):

src/completion-listener.ts
odeadglaz: Got ya, make sense 👍

src/completion-listener.ts
odeadglaz: Yea good point 👍 It would mean we will change the `CommandInfo` interface to include it as well 👌

odeadglaz:
@gustavohenke Not sure why I'm getting inconsistent behavior in tests, did you encounter this kind of behavior in the past? (locally over my mac I can't seem to reproduce it)

src/completion-listener.ts
gustavohenke: Heya, I've fixed this myself and added an extra test too.

You were right, adding to `CommandInfo` is not a good idea, but so wasn't right to pass the state in the close event since that's pretty static. Can check `commands` variable just outside the loop, though.

gustavohenke (approved):

gustavohenke:
Hey! This is now available in v9.0.0.

PR #479, opened 2024-05-26
Bring spawn-command into concurrently

### LINKED ISSUE #407: Drop dependency on `spawn-command`
We recently started using https://socket.dev/ as part of our dependencies maintenance and it flagged `spawn-command` as problematic due to it being unmaintained: https://socket.dev/npm/package/spawn-command

The last release was 8 years ago, so there's probably other solutions out there or even can be written with native Node.js nowadays.

### LINKED ISSUE #477: Deprecation warning with Node 22
After upgrading to Node 22 we get a deprecation warning `DEP0060` caused by the dependency `spawn-command`:

```
at module.exports (...\node_modules\spawn-command\lib\spawn-command.js:9)
at Proxy.start (...\node_modules\concurrently\dist\src\command.js:55:28)
...
```
The code that causes that is [this line](https://github.com/mmalecki/spawn-command/blob/master/lib/spawn-command.js#L9)

The [fix](https://nodejs.org/api/deprecations.html#DEP0060) looks trivial but `spawn-command` doesn't seem to be developed anymore. Maybe the functionality could be integrated in `concurrently`.

gustavohenke:
Yay we no longer have a 10+ years old unmaintained dependency!

Fixes #407
Fixes #477

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,6 @@ tired of opening terminals and made **concurrently**.
 - Cross platform (including Windows)
 - Output is easy to follow with prefixes
 - With `--kill-others` switch, all commands are killed if one dies
-- Spawns commands with [spawn-command](https://github.com/mmalecki/spawn-command)
 
 ## Installation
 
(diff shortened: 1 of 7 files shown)


theonlypwner:
mmalecki/spawn-command#9 was fixed, so all that really needed to be done was to bump `spawn-command` to `0.0.3`.

gustavohenke:
Hey. This is now fixed in v9.0.0.

PR #486, opened 2024-06-13
Make `--raw` compatible with `--hide`

### LINKED ISSUE #394: --hide param doesn't work with --raw output
I have tried: 
``` bash
concurrently --kill-others "npm run build:watch" "npm run start:pm2" -r --hide "0"
# OR 
concurrently --kill-others "npm run build:watch" "npm run start:pm2" -r --hide 0
```

But hiding works only when I remove the `-r` (raw) param.

Sorry if this is intended behavior.

PoQuatre:
This PR closes #394

### DIFF
--- a/bin/concurrently.spec.ts
+++ b/bin/concurrently.spec.ts
@@ -234,6 +234,22 @@ describe('--hide', () => {
         expect(lines).toContainEqual(expect.stringContaining('foo'));
         expect(lines).not.toContainEqual(expect.stringContaining('bar'));
     });
+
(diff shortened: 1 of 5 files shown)


PoQuatre:
I did some tests on my side with a package I'm working on, and it seems to work without issues, although I'm not 100% sure if I haven't missed some edge cases as I am not super familiar with the tool and the codebase

gustavohenke (approved):
Hey! I don't think there's any edge case here, just some optimisations and nitpicks, but I can apply those after merging your contribution in.

Issue #487, opened 2024-06-14
escaping pass-through arguments is too eager


soletan:
## Background

When trying to use **concurrently** with `--passthrough-arguments` option, any passed arguments get escaped too eagerly. 

The issue has been observed while developing an application when trying to concurrently run a local web server and [gherkin-testcafe](https://www.npmjs.com/package/gherkin-testcafe#cli-usage) for testing with the latter requring tags to start with @ character.

Since that application can't be shared here for legal reasons I've tried a simpler scenario which is reproducing the issue, too.

## Scenario

- windows platform
- npm 10.5.2
- node 20.13.1
- concurrently 8.2.2
- files:
  - package.json contains scripts:
    ```json
    {"scripts": {
        "foo": "concurrently -k -n \"bar,baz\" -P npm:foo:bar \"npm:foo:baz -- {@}\"",
        "foo:bar": "node block.js",
        "foo:baz": "node log.js",
    }}
    ```
  - file block.js looks like this:
    ```javascript
    setTimeout(() => console.log("delayed"), 1000);
    ```
  - file log.js looks like this:
    ```javascript
    console.dir(process.argv);
    ```

### invoke without arguments

```bash
$ npm run foo

> logistics-client@0.0.0 foo
> concurrently -k -n "bar,baz" -P npm:foo:bar "npm:foo:baz -- {@}"

[bar] 
[bar] > logistics-client@0.0.0 foo:bar
[bar] > node block.js
[bar]
[baz]
[baz] > logistics-client@0.0.0 foo:baz
[baz] > node log.js
[baz]
[baz] [
[baz]   'C:\\Program Files\\nodejs\\node.exe',
[baz]   'C:\\Users\\cepharum\\PhpstormProjects\\UID\\Logistics\\WebClient\\Source\\UiNew\\log.js'
[baz] ]
[baz] npm run foo:baz --  exited with code 0
--> Sending SIGTERM to other processes..
[bar] npm run foo:bar exited with code 1
```
looks good ... setup is working as expected

### invoke with arguments to pass

```bash
$ npm run foo -- -- --tags=@foo

> logistics-client@0.0.0 foo
> concurrently -k -n "bar,baz" -P npm:foo:bar "npm:foo:baz -- {@}" -- --tags=@foo

[bar]
[bar] > logistics-client@0.0.0 foo:bar
[bar] > node block.js
[bar]
[baz]
[baz] > logistics-client@0.0.0 foo:baz
[...]

PR #491, opened 2024-06-22
switch quoting library

soletan:
fixing #487

The PR is replacing the generic approach in shell-quote library with a different one based on character-by-character tests I've made resulting in a different quoting library.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,11 +55,11 @@
   "author": "Kimmo Brunfeldt",
   "license": "MIT",
   "dependencies": {
+    "@cepharum/quoting-db": "^1.2.0",
     "chalk": "^4.1.2",
     "date-fns": "^2.30.0",
     "lodash": "^4.17.21",
     "rxjs": "^7.8.1",
-    "shell-quote": "^1.8.1",
     "supports-color": "^8.1.1",
     "tree-kill": "^1.2.2",
     "yargs": "^17.7.2"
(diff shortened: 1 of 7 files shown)


soletan:
Hm, not sure about the open linter errors for running ESLint on my working copy presents that warning, only. In case it is fixing something locally, this does not result in a change accepted for commission by git CLI tool.

pnpm-lock.yaml
gustavohenke, line 1: Seems like a lockfile version bump here.
I'll try updating this on main first to decrease the LOC in this PR.

package.json
```diff
@@ -55,11 +55,11 @@
   "author": "Kimmo Brunfeldt",
   "license": "MIT",
   "dependencies": {
+    "@cepharum/quoting-db": "^1.2.0",
```
gustavohenke, line 58: Seems like a fresh new package, would you be keen on hosting and maintaining it from the open-cli-tools org? Seems it'd fit perfectly ;)

src/command-parser/expand-arguments.ts
```diff
@@ -10,6 +10,8 @@ export class ExpandArguments implements CommandParser {
     constructor(private readonly additionalArguments: string[]) {}
 
     parse(commandInfo: CommandInfo) {
+        const configuration = getShellConfigurationSync();
```
gustavohenke, line 13: Looked at the source of this function briefly, seems it defaults to the value of `COMSPEC`.
Since cmd.exe is hardcoded for windows (source below), should the same be done here?

https://github.com/open-cli-tools/concurrently/blob/e52d984fa179e90b2d8f3548012e38987500a8a3/src/spawn.ts#L17-L18

gustavohenke (changes requested):
> not sure about the open linter errors

Maybe when updating pnpm lockfile eslint or prettier versions got updated, which now cause those?

PR #495, opened 2024-08-04
Prefix padding

### LINKED ISSUE #166: Options to make sure names are the same length
Would you accept a PR that adds a `--min-prefix-length` options?

When using wildcards command, it's harder to make sure the name are of the same size (for aesthetic reasons)

ei.

```
[babel] Successfully compiled 26 files with Babel.
[webpack] Compiled successfully.
```

Would become (using `--min-prefix-length 7`)

```
[babel  ] Successfully compiled 26 files with Babel.
[webpack] Compiled successfully.
```

### LINKED ISSUE #417: Add a --fix-name-length option to align command output
When using concurrently, providing custom names with the `-n` or `--names` option can cause output to become misaligned and difficult to read if the names have different lengths. This can be particularly problematic when the output is lengthy and spans multiple lines.

To address this issue, I propose adding a new `--fix-name-length` option that allows users to specify a fixed length for the command names. When this option is used, the command names will be padded with spaces or truncated as necessary to ensure that they all have the same length, and the output will be properly aligned.

For example, if a user runs the following command:

```
concurrently -n "API, TypeScript, VeryLong" --fix-name-length\
    "npm run api" "npm run typescript" "npm run somethingverylong"
```

The output would be aligned like this:

```
API        | output from npm run api
TypeScript | output from npm run typescript
VeryLong   | output from npm run somethingverylong
```

This would make it much easier to read and understand the output from each command, especially in cases where the output spans multiple lines.

I believe that adding a `--fix-name-length` option would be a valuable addition to concurrently, as it would improve the user experience for those who use custom names for their commands.

gustavohenke:
Added a `--pad-prefix` flag which makes all prefixes the same length, padding shorter ones with spaces.

Example:

```
$ concurrently -n foo,barbaz --pad-prefix "echo foo" "echo bar"
[foo   ] foo
[foo   ] echo foo exited with code 0
[barbaz] bar
[barbaz] echo bar exited with code 0
```

**Noteworthy:** 
Although most prefixes have a stable length, PIDs can become longer if the command restarts, e.g. it was originally 9 but after a restart it's 10.
Obviously we can't rewrite the past, so new log entries from that respawn onwards will be padded appropriately.

Closes #166 
Closes #417

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -14,6 +14,12 @@ export class Logger {
     private readonly timestampFormat: string;
 
     /**
+     * How many characters should a prefix have.
+     * Prefixes shorter than this will be padded with spaces to the right.
+     */
+    private prefixLength = 0;
+
+    /**
      * Last character emitted.
      * If `undefined`, then nothing has been logged yet.
      */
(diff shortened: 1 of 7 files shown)


src/logger.ts
```diff
@@ -84,33 +90,51 @@ export class Logger {
 
     private getPrefixesFor(command: Command): Record<string, string> {
         return {
-            pid: String(command.pid),
+            pid: command.pid != null ? String(command.pid) : '',
```
gustavohenke, line 93: So that it's not `'undefined'` when there's limited concurrency, which could be longer than the actual PID

gustavohenke:
```suggestion
            // When there's limited concurrency, the PID might not be immediately available,
            // so avoid the string 'undefined' from becoming a prefix
            pid: command.pid != null ? String(command.pid) : '',
```

PR #498, opened 2024-08-13
Inter-Process Communication

gustavohenke:
This PR implements basic inter-process communication (IPC), which is an API-only feature - not available through CLI.

`Command` has the following new fields/methods:
- `send`: to send messages
- `messages.incoming`: to listen for received messages
- `messages.outgoing`: to listen for sent messages

The caller must set the `ipc` property when calling `concurrently()`. It sets the number of the file descriptor on which to set up the IPC channel.

Example usage:

```js
const { commands } = concurrently([{ command: 'echo foo', ipc: 3 }]);
commands[0].send({ hello: 'world' })
commands[0].messages.incoming.subscribe(({ message }) => {
	console.log('command 0 sent this:', message);
});
```

With this PR I also realise that the README is no longer a good option for documentation.
We need to expand it into a website...

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,7 +318,7 @@ For more details, visit https://github.com/open-cli-tools/concurrently
 ### `concurrently(commands[, options])`
 
 - `commands`: an array of either strings (containing the commands to run) or objects
-  with the shape `{ command, name, prefixColor, env, cwd }`.
+  with the shape `{ command, name, prefixColor, env, cwd, ipc }`.
 
 - `options` (optional): an object containing any of the below:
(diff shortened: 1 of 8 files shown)


gustavohenke:
🚢 This is now available in v9.0.0!

studiosciences:
This is great! Thx @gustavohenke.

I created a simple "waitForStartMessage.js" script, so that I could delay a process until our Orval code generation process is setup.

```
    {
      command: `node ${__dirname}/waitForStartMessage.js && yarn start`,
      name: "ui",
      ipc: 3,
    },
```

PR #508, opened 2024-10-11
Add support for Deno shortcuts and wildcards 

mahtaran:
With the release of Deno 2, I figured I'd have a look if the amazing shortcut and wildcard features of this tool supported it. Sadly, they didn't. Luckily, adding it wasn't too difficult. One thing to note is that Deno uses its own `deno.json` file with what it calls 'tasks', but can also fall back on Node's `scripts` directive in the `package.json`.

For a cleaner reviewing process, I recommend looking at the individual commits, as it doesn't quite seem to understand I renamed some files.

### DIFF
--- a/src/command-parser/expand-npm-shortcut.ts
+++ b/src/command-parser/expand-npm-shortcut.ts
@@ -2,21 +2,38 @@ import { CommandInfo } from '../command';
 import { CommandParser } from './command-parser';
 
 /**
- * Expands commands prefixed with `node:`, `npm:`, `yarn:`, `pnpm:`, or `bun:` into the full version `npm run <command>` and so on.
+ * Expands shortcuts according to the following table:
+ *
+ * | Syntax          | Expands to            |
+ * | --------------- | --------------------- |
+ * | `npm:<script>`  | `npm run <script>`    |
+ * | `pnpm:<script>` | `pnpm run <script>`   |
+ * | `yarn:<script>` | `yarn run <script>`   |
+ * | `bun:<script>`  | `bun run <script>`    |
+ * | `node:<script>` | `node --run <script>` |
+ * | `deno:<script>` | `deno task <script>`  |
  */
 export class ExpandNpmShortcut implements CommandParser {
     parse(commandInfo: CommandInfo) {
-        const [, npmCmd, cmdName, args] =
-            commandInfo.command.match(/^(node|npm|yarn|pnpm|bun):(\S+)(.*)/) || [];
-        if (!cmdName) {
+        const [, prefix, script, args] =
+            /^(npm|yarn|pnpm|bun|node|deno):(\S+)(.*)/.exec(commandInfo.command) || [];
+        if (!script) {
             return commandInfo;
         }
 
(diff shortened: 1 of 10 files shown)


src/command-parser/expand-wildcard.ts
```diff
@@ -0,0 +1,109 @@
+import fs from 'fs';
+import _ from 'lodash';
+
+import { CommandInfo } from '../command';
+import { CommandParser } from './command-parser';
+
+const OMISSION = /\(!([^)]+)\)/;
```
igorjs, line 7: Would you mind adding a comment explaining what this Regexp does?
It isn't clear its intent unless you read the rest of the code and, even then, it's a bit confused.

Thanks in advance.

mahtaran: Yeah, I was confused when I read it at first as well. Good catch, just added a comment. Do you think that clarifies it enough?

igorjs: Thank you.

I was just wondering why we'd need to look for `)` in the script command?!

I guess most of the people (myself included) don't know we could use something like this `deno task lint:*(!fix)` as a valid command.

Perhaps, adding an example in the comment would help as well.

mahtaran: This behaviour is [currently already documented](https://github.com/open-cli-tools/concurrently/blob/main/docs/cli/shortcuts.md?plain=1#L61), so I feel this is a bit out of scope of this pull request?

igorjs (approved):

src/command-parser/expand-npm-shortcut.ts
```diff
@@ -2,21 +2,38 @@ import { CommandInfo } from '../command';
...
+ * | `node:<script>` | `node --run <script>` |
+ * | `deno:<script>` | `deno task <script>`  |
  */
 export class ExpandNpmShortcut implements CommandParser {
     parse(commandInfo: CommandInfo) {
-        const [, npmCmd, cmdName, args] =
-            commandInfo.command.match(/^(node|npm|yarn|pnpm|bun):(\S+)(.*)/) || [];
-        if (!cmdName) {
+        const [, prefix, script, args] =
+            /^(npm|yarn|pnpm|bun|node|deno):(\S+)(.*)/.exec(commandInfo.command) || [];
```
gustavohenke: Out of curiosity, was there any advantage of changing from `string.match` to `regexp.exec()` here?
AFAICT both are generally equivalent, but maybe I'm missing something.

No dramas if it's just your preference :)

gustavohenke (approved):
Great addition, thank you!

src/command-parser/expand-npm-shortcut.ts
mahtaran: Mostly personal preference, yeah. From what I've heard, `String.match()` and `RegExp.exec()` function identically to each other if the global flag is not set, in fact, `String.match()` uses `RegExp.exec()` under the hood, so using the latter carries a very minor performance benefit with it.

PR #509, opened 2024-10-20
Update readme usage instructions for quotes

victorhcb:
This discussion came up in a Brazilian project (TabNews) when a [Pull Request](https://github.com/filipedeschamps/tabnews.com.br/pull/592) revealed a compatibility issue with Windows due to the use of single quotes.

The idea of this PR is to make it clear in the documentation that using double quotes is the way to go if you want to avoid compatibility problems across different operating systems.

It’s a small change, but it could save users from wasting minutes—or even hours—trying to figure out what went wrong!

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,10 @@ tired of opening terminals and made **concurrently**.
 
 The tool is written in Node.js, but you can use it to run **any** commands.
 
-Remember to surround separate commands with quotes:
+Remember to surround separate commands with **double quotes**:
+
+> **Note**
+> The use of double quotes instead of single quotes is important to maintain compatibility between different operating systems.
 
 ```bash
 concurrently "command1 arg" "command2 arg"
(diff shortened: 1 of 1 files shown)


gustavohenke (commented):
Hey. I'm not sure if it's relevant to highlight it like that, as it's not concurrently specific, but rather how the shell parses inputs.

If we want to be thorough with the recommendations, then I'd argue that someone blindly changing their commands from single to double quotes could also face unintended side effects if they look like they have an env variable, e.g.

```
$ conc 'echo $foo'
[0]
[0] echo $foo exited with code 0
$ conc "echo $foo"
[0]
[0] echo  exited with code 0
```

May I suggest that maybe we change the note to say that the user should pick the right quotes for their use case?
Can highlight how Windows will only work with double ones, and that env vars in unix might be parsed eagerly depending on the style.

Issue #482, opened 2024-05-28
Making concurrently a purely ESM package


aqeelat:
Since concurrently is mostly run via CLI, what is stopping it from being converted to ESM? Are there any implications that are slipping my mind?

I'm willing to try converting it if the change would be accepted.

gustavohenke:
Hey! Nothing _really_ stopping it, it's just that I wouldn't like to exclude some CJS users of the API.
I don't think there's any benefit of doing this change either, other than using newer versions of e.g. chalk?

Also a fun read:
https://www.reddit.com/r/node/comments/14rg9ym/esm_not_gaining_traction_in_backend_node/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

Seems like the numbers are a bit dated, but chalk ESM-only is still 10x less popular than its previous, CJS version. [chalk on npm](https://www.npmjs.com/package/chalk?activeTab=versions)

gustavohenke:
I'm closing this issue for now, but happy to read what others think.

PR #523, opened 2024-12-26
fix: support Deno's JSON with comments configuration

mahtaran:
Deno supports JSONC (JSON with comments) as project configuration, in both `deno.json` and `deno.jsonc`.

Had to manually copy the dependency because it wouldn't work in a CommonJS project. Argument in favour of #482.


gustavohenke (approved):
Thanks!

gustavohenke:
Published as v9.1.1.

PR #526, opened 2025-01-16
Add new --matrix option to multiply commands

### LINKED ISSUE #517: Add CLI support for command templates/matrix commands
It would be useful to be able to generate the commands dynamical based an environment variable. For example:

```sh
concurrently --matrix "A B C" "echo {}"
```

Which would be equivalent to typing:

```sh
concurrently "echo A" "echo B" "echo C"
```

Or with multiple values you could do:

```sh
concurrently --matrix "A B C" --matrix "1 2 3" "echo {1}{2}"
```

Which would be equivalent to writing:

```sh
concurrently "echo A1" "echo "A2" "echo A3" "echo B1" "echo B2" "echo B3" "echo C1" "echo C2" "echo C3"
```

This would be useful for common repetitive tasks that can happen in parallel. For example, let's say I want to download some docker image on a few different servers. I could do:

```sh
SERVERS="server1 server2"
IMAGES="python:3.11 node:20"

concurrently --group --matrix "$SERVERS" --matrix "$IMAGES" "ssh {1} 'docker pull {2}'"
```

This would help making `concurrently` a drop-in replacement for bash for-loops; the above would replace

```sh
for server in $SERVERS; do
    for image in $IMAGES; do
        ssh $server "docker pull $image"
    done
done
```

It would also make `concurrently` a drop-in for a lot of gnu `parallel` usages, eg:

```sh
parallel --quote ssh {1} "docker pull {2}" ::: $SERVERS ::: $IMAGES
```

Another example: processing a lot files in parallel, eg less compiling a number of files:

```sh
concurrently --matrix "$(ls *.less)" --group "lessc {}"
```

cdrini:
Closes #517 

Sample runs:

Single dimension:
```sh
$ dist/bin/concurrently.js --group --matrix "os:windows linux" 'echo {M:os}'
[0] windows
[0] echo windows exited with code 0
[1] linux
[1] echo linux exited with code 0
```

Multiple dimensions:
```sh
$ dist/bin/concurrently.js --group -M "os:windows linux" -M "version:10 12 14" 'echo {M:os} (version {M:version})'
[0] windows (version 10)
[0] echo windows (version 10) exited with code 0
[1] windows (version 12)
[1] echo windows (version 12) exited with code 0
[2] windows (version 14)
[2] echo windows (version 14) exited with code 0
[3] linux (version 10)
[3] echo linux (version 10) exited with code 0
[4] linux (version 12)
[4] echo linux (version 12) exited with code 0
[5] linux (version 14)
[5] echo linux (version 14) exited with code 0
```

With environment variables:
```sh
$ SERVERS="$(echo ol-web{0..2})"
$ IMAGES="python:3.12-slim python:3.13-slim"
$ dist/bin/concurrently.js --group -r -M "server:$SERVERS" -M "image:$IMAGES" 'echo ssh {M:server} docker image pull {M:image}'
ssh ol-web0 docker image pull python\:3.12-slim
ssh ol-web0 docker image pull python\:3.13-slim
ssh ol-web1 docker image pull python\:3.12-slim
ssh ol-web1 docker image pull python\:3.13-slim
ssh ol-web2 docker image pull python\:3.12-slim
ssh ol-web2 docker image pull python\:3.13-slim
ssh ol-web1 docker image pull python\:3.12-slim
ssh ol-web1 docker image pull python\:3.13-slim
ssh ol-web2 docker image pull python\:3.12-slim
ssh ol-web2 docker image pull python\:3.13-slim
ssh ol-web2 docker image pull python\:3.12-slim
ssh ol-web2 docker image pull python\:3.13-slim
```


src/index.ts
```diff
@@ -103,6 +103,11 @@ export type ConcurrentlyOptions = Omit<BaseConcurrentlyOptions, 'abortSignal' |
      * If not defined, no argument replacing will happen.
      */
     additionalArguments?: string[];
+
+    /**
+     * This command should be run multiple times, for each of the provided matrices.
```
gustavohenke, line 108: Can you please update this comment? It looks like it's assuming some context from elsewhere.

src/concurrently.ts
```diff
@@ -179,6 +185,10 @@ export function concurrently(
         new ExpandWildcard(),
     ];
 
+    if (options.matrices?.length) {
```
gustavohenke, line 188: Looks like matrices have preference over passthrough arguments using the same `{number}` syntax. Was it intended?

It needs to be called out in the help and in the docs.

gustavohenke (changes requested):
Thanks for taking this up!
I made several suggestions and asked a few questions in this pass.

I'll update [the docs](https://github.com/open-cli-tools/concurrently/blob/main/docs/cli/passthrough-arguments.md) to mention matrices as well in a follow-up, unless you feel like doing it while keeping it on brand?

paescuj:
@cdrini Wanna take this up again?

cdrini:
Thank you for rebasing, @paescuj ! And thank you for the code review @gustavohenke . Apologies for the delay was a bit stuck on the right approach. Taking another look today.

src/command-parser/expand-matrices.ts
```diff
@@ -0,0 +1,62 @@
...
+import { CommandParser } from './command-parser';
+
+/**
+ * Replace placeholders with new commands for each combination of matrices.
+ */
+export class ExpandMatrices implements CommandParser {
+    private _bindings: string[][];
+
+    constructor(private readonly matrices: readonly string[][]) {
+        this.matrices = matrices;
```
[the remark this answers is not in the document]
cdrini: I think I cleaned all these up? I'm not too familiar with the best practices for TS for fields like these, let me know if there's something missing!

cdrini:
Ok! I updated this up with the comments. I decided to tweak the interface slightly to allow/enforce _named_ matrix dimensions, and a `M:` prefix for the matrix placeholder in the command (eg `--matrix "os:windows linux" 'echo {M:os}'`). This not only fixes the collisions issue with the placeholder for pass-through arguments, but also avoids an issue I've noticed with GNU parallel, which is that it very quickly becomes unreadable with very many eg `{1}` `{3}` `{2}`. It also avoids having to have extra options for names (something GNU parallel supports IIRC), which I think fragments the interface a bit. Now there's one interface that kind of handles it all. Although at the cost of being a bit more verbose.

I updated the description with the new interface. Would love to know your thoughts!

Note: I also fixed a naming error I'd made in the code; there aren't multiple `matrices`, there is just one `matrix`!

~Also if anyone has any thoughts on why that mac test is failing... 🤔~ Thank you to whoever I assume re-ran it.

Issue #497, opened 2024-08-10
Make command more readable


johndeighan:
Isn't it possible to allow something like this in my package.json file:

		"dev": "concurrently 'llb -w' 'vite -c src/vite.config.js src' 'promptme -w'"

i.e. use single quotes inside the concurrently command string. This would be much easier to read and write instead of having to use backslash escaped double quotes. JSON requires strings to be delimited by double quotes, but there's no such requirement when it comes to characters inside a string.

I tried it, but it didn't work ;-(

gustavohenke:
Hey @johndeighan, that's actually valid, but not on Windows, AFAIK: https://superuser.com/q/324278/200254
Are you using Windows, by any chance?

PR #537, opened 2025-04-03
docs: nicer quotes

IsaacLeeWebDev:
It's generally a good practice to prefer single-quotes in shell script strings unless you plan on interpolating variables. 

It's purely stylistic here (even if you disregard that these are just markdown files 😅), but there's something about shell code that looks safer at first-glance. It's sorta like how a SQL query string _can_ be safe when concatenating another string directly onto it, but it doesn't _look_ as safe as one that escapes inputs.

Plus, given that the NPM scripts are already inside JSON strings, getting rid of escaped characters makes it look neater! :3

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ The tool is written in Node.js, but you can use it to run **any** commands.
 Remember to surround separate commands with quotes:
 
 ```bash
-concurrently "command1 arg" "command2 arg"
+concurrently 'command1 arg' 'command2 arg'
 ```
 
 Otherwise **concurrently** would try to run 4 separate commands:
@@ -70,7 +70,7 @@ Otherwise **concurrently** would try to run 4 separate commands:
 In package.json, escape quotes:
 
 ```bash
-"start": "concurrently \"command1 arg\" \"command2 arg\""
+"start": "concurrently 'command1 arg' 'command2 arg'"
 ```
 
 You can always check concurrently's flag list by running `concurrently --help`.
--- a/docs/cli/input-handling.md
+++ b/docs/cli/input-handling.md
@@ -4,7 +4,7 @@ By default, concurrently doesn't send input to any commands it spawns.<br/>
 In the below example, typing `rs` to manually restart [nodemon](https://nodemon.io/) does nothing:
 
 ```bash
-$ concurrently "nodemon" "npm run watch-js"
+$ concurrently 'nodemon' 'npm run watch-js'
 rs
 ```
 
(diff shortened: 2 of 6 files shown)


IsaacLeeWebDev:
I love this tool SO much, and use it all the time! Thanks for all the hard work!

IsaacLeeWebDev:
Also, I made an exception here:

https://github.com/IsaacLeeWebDev/concurrently/blob/3c428271e00198f13791a30e154f2eccf8d33ce4/docs/cli/configuration.md#L10

Just 'cause the way to escape a single-quote is a bit tricky to read, and at the end of the day, docs are meant to be read!

gustavohenke (commented):
Fair change, I think - but looks the quotes thing has bitten some people before: #509.
What are your thoughts?

I don't mind it either way.

IsaacLeeWebDev:
Man, it's a really tough choice.

Because `concurrently` is a lot like unix's `eval` in that it intentionally executes string arguments as code, and because there are widely used alternatives to Command Prompt on Windows that don't have the problem in #509 such as PowerShell, I think I personally lean more towards an annoyingly harsh (and potentially untimely and costly) lesson about portability. 

While the odds of someone passing potentially unsanitized user input to `concurrently` before they know how quotes work because they copy-pasted from `README.md` seem pretty darn remote, if and when it happens, boy would that be a harsh lesson. 

In contrast, learning to make commands you find in docs portable to your command interpreter of choice seems almost inevitable. 

I think preserving the "copy-pasteability" of your docs is an very worthwhile goal, and command portability seems like a wide-spread accessibility issue that the world of software is wrought with. 

I think I still lean more towards single-quotes, but it's really just not cut and dry to me, and if nothing changed, I'd get it. I think it just sorta comes down to the values of the maintainers, and no matter what, I definitely empathize with anyone who gets hung up on Windows vs. DOS v.s Unix. 

I wonder what other CLI tools do about this in their docs! Maybe it's worth listing two versions of the command in the README for Command Prompt users?

gustavohenke (approved):
Thanks for sharing your thoughts!

>  comes down to the values of the maintainers

> I guess there's some marginal amount of value in adopting a style in documentation that errs a bit more on the secure/paranoid side

These have echoed with me because I've been writing the examples somewhat mindlessly based on what I do myself.
However, being intentional on what to use means adopting a style, which helps reducing messy examples each using their own style going forward.

And as you pointed out AI, these days being intentional might even influence these tools to write examples in using the safer default of single quotes.

I also came across #497 that I didn't remember about, and being intentional about the use of single quotes might help a few users.
Besides, probably not a lot of people use cmd these days, and powershell seems to support single quotes.

> Maybe it's worth listing two versions of the command in the README for Command Prompt users?

Maybe, if there was a documentation website that could hide the duplicated examples.
GitHub can't do that yet, though, so it becomes noise.

PR #549, opened 2025-06-15
chore: update jest from v29 to v30

noritaka1166:
update jest from v29 to v30
- <https://jestjs.io/docs/upgrading-to-jest30>
- <https://github.com/jestjs/jest/blob/main/CHANGELOG.md>


paescuj:
Thanks!

PR #554, opened 2025-07-06
Replace jest with vitest

gustavohenke:
I overall like vitest more, however, the main motivation to replace it is that it'll have better ESM support 🙂


gustavohenke:
@paescuj I see you've been reviewing a few PRs, wanna give this one a sanity check?

PR #565, opened 2025-08-25
Drop support for Node.js 18, add support for Node.js 24

paescuj:
Node.js 18: Active Support ended on 18 Oct 2023, Security Support ended on 30 Apr 2025.

Node.js 24: Upcoming LTS over the next months.

See also https://nodejs.org/en/about/previous-releases#release-schedule.

---

This also resolves the following Vitest message seen on Node.js 18:
> The CJS build of Vite's Node API is deprecated. See https://vite.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.


gustavohenke (approved):
Nice, thanks!

PR #569, opened 2025-10-02
Bundle the code

paescuj:


gustavohenke (commented):
I'll be a hard sell on bundling.

But looks like you're including more than just bundling here?

PR #574, opened 2025-10-16
Correct modifier application description in prefixing.md

CodeF53:
Modifiers are appended to the colors `#e700ff.bold` prepending would be like this `bold.#e700ff`


mochienya (approved):

paescuj:
Thanks!

PR #581, opened 2026-03-06
feat: enable automatic prefix colors by default

GermanJablo:
## Summary

- Changes the default `--prefix-colors` from `reset` (no colors) to `auto`, so each command gets a distinct prefix color out of the box.
- Sets the auto-cycling palette to **cyan, magenta, green, yellow, blue**.
- Fixes the logger's `colorText` fallback to handle non-Chalk values in `defaults.prefixColors` gracefully.

## Motivation

Running `concurrently "cmd1" "cmd2" "cmd3"` today produces uncolored prefixes — all `[0]`, `[1]`, `[2]` look the same unless the user explicitly passes `-c auto` or a color list. Colored prefixes are one of the most useful visual cues when scanning interleaved output, yet they require opting in.

This PR makes colors the default. Users who prefer uncolored output can pass `-c reset`.

### Color choice rationale

The previous `ACCEPTABLE_CONSOLE_COLORS` list had 16 entries including colors that are problematic as defaults:

- **`white` / `grey`** blend into the terminal's default text color on most themes, looking effectively uncolored.
- **`red`** is universally associated with errors; using it for normal output creates a false sense of alarm.
- **`bg*` colors** (background-colored text) are visually heavy and distracting for everyday use.

The chosen 5-color palette avoids all of these: each color is clearly distinguishable, reads well on both dark and light backgrounds, and carries no semantic baggage. This is the same set of colors used by turborepo, which served as inspiration for the selection.

### Default pool vs allowed colors

`ACCEPTABLE_CONSOLE_COLORS` controls what `auto` picks from — it is **not** a whitelist. Manually specified colors via `--prefix-colors` remain completely unrestricted: any valid Chalk color name (`red`, `bgCyan`, `greenBright`, …), hex value (`#23de43`), or modifier (`bold`, `inverse`, …) works exactly as before. This distinction is now documented in the source.

## Changes

| File | What changed |
|------|-------------|
| `lib/defaults.ts` | `prefixColors`: `'reset'` → `'auto'` |
| `lib/prefix-color-selector.ts` | `ACCEPTABLE_CONSOLE_COLORS` → 5-color palette with updated JSDoc |
| `lib/logger.ts` | `colorText` fallback: `?? this.chalk.reset` for safety |
| `lib/prefix-color-selector.spec.ts` | New test asserting palette matches the expected 5 colors |
| `docs/cli/prefixing.md` | Updated to describe the new default behavior |

## Testing

```bash
pnpm test -- --run lib/prefix-color-selector.spec.ts
```

To see the colors locally:

```bash
pnpm build && node dist/bin/index.js "echo cmd1" "echo cmd2" "echo cmd3" "echo cmd4" "echo cmd5" "echo cmd6" "echo cmd7"
```

Commands 6 and 7 cycle back to cyan and magenta. Manual colors still work:

```bash
node dist/bin/index.js -c "red,bgCyan,white" "echo a" "echo b" "echo c"
```

All 583 tests pass (582 existing + 1 new). Lint and type checks clean.

### Screenshots to visualize the colors

image

Example from the repo:

image

### DIFF
--- a/lib/prefix-color-selector.spec.ts
+++ b/lib/prefix-color-selector.spec.ts
@@ -181,7 +181,7 @@ describe('#ACCEPTABLE_CONSOLE_COLORS', () => {
         expect(PrefixColorSelector.ACCEPTABLE_CONSOLE_COLORS.length).toBeGreaterThan(1);
     });
 
-    it('uses the same color order as turborepo', () => {
+    it('uses the same colors and order as turborepo', () => {
         expect(PrefixColorSelector.ACCEPTABLE_CONSOLE_COLORS).toEqual([
             'cyan',
             'magenta',
--- a/lib/prefix-color-selector.spec.ts
+++ b/lib/prefix-color-selector.spec.ts
@@ -180,4 +180,14 @@ describe('#ACCEPTABLE_CONSOLE_COLORS', () => {
         //     always has more than one entry, which is what we enforce via this test
         expect(PrefixColorSelector.ACCEPTABLE_CONSOLE_COLORS.length).toBeGreaterThan(1);
     });
+
+    it('uses the same color order as turborepo', () => {
+        expect(PrefixColorSelector.ACCEPTABLE_CONSOLE_COLORS).toEqual([
+            'cyan',
+            'magenta',
+            'green',
+            'yellow',
+            'blue',
+        ]);
+    });
 });
--- a/lib/prefix-color-selector.spec.ts
+++ b/lib/prefix-color-selector.spec.ts
@@ -181,7 +181,7 @@ describe('#ACCEPTABLE_CONSOLE_COLORS', () => {
         expect(PrefixColorSelector.ACCEPTABLE_CONSOLE_COLORS.length).toBeGreaterThan(1);
     });
 
-    it('uses the same colors and order as turborepo', () => {
+    it('only includes colors that are visually distinct, semantically neutral, and lightweight', () => {
         expect(PrefixColorSelector.ACCEPTABLE_CONSOLE_COLORS).toEqual([
             'cyan',
             'magenta',
--- a/lib/prefix-color-selector.ts
+++ b/lib/prefix-color-selector.ts
@@ -63,33 +63,19 @@ export class PrefixColorSelector {
         this.colorGenerator = createColorGenerator(normalizedColors);
     }
 
-    /** A list of colors that are readable in a terminal. */
+    /**
+     * Colors used by `auto` selection and default cycling.
+     *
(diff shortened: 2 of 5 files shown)


lib/prefix-color-selector.ts
```diff
@@ -63,33 +63,19 @@ export class PrefixColorSelector {
         this.colorGenerator = createColorGenerator(normalizedColors);
     }
 
-    /** A list of colors that are readable in a terminal. */
+    /**
+     * Colors used by `auto` selection and default cycling.
+     *
+     * Matches turborepo's palette: cyan, magenta, green, yellow, blue.
```
gustavohenke: Unnecessary to call out the similarity to turborepo and the colour list - both can go out of the date.

Then, without them, the rest of the comment needs to be adjusted to keep the notes on visual distinction, semantic, and contrast.

lib/prefix-color-selector.spec.ts
```diff
@@ -180,4 +180,14 @@ describe('#ACCEPTABLE_CONSOLE_COLORS', () => {
         //     always has more than one entry, which is what we enforce via this test
         expect(PrefixColorSelector.ACCEPTABLE_CONSOLE_COLORS.length).toBeGreaterThan(1);
     });
+
+    it('uses the same colors and order as turborepo', () => {
```
gustavohenke: Instead of comparing ourselves to turborepo, I think this test could instead guarantee the main properties described in your PR - contrast, semantics, and not heavy.
So it checks that the colours are not white/grey/black/red/background ones.

WDYT?

gustavohenke (changes requested):
Absolutely on board with most changes in this PR.

Change from `reset` to `auto` default could be backed by analytics, but we don't have these at the moment.
Still, I'm inclined to accept it as a breaking change, it's just curious how it was never requested before.

lib/prefix-color-selector.spec.ts
GermanJablo: I totally agree! I told my AI the exact same thing, and it corrected it in the PR description but forgot to do it here. Sorry about that!

GermanJablo:
Hi @gustavohenke, I addressed your comments a couple of weeks ago, but I'm not sure if you saw it. Let me know if this is good to go.

GermanJablo:
Hi @gustavohenke, sorry to ping you again. I just wanted to double-check if you saw this.

GermanJablo:
Friendly reminder 🙏😅

gustavohenke (approved):
Thanks very much! And sorry for taking such a long time to get back to your PR.

PR #585, opened 2026-04-17
fix: scope quote normalization to CLI input

### LINKED ISSUE #582: StripQuotes regex incorrectly strips outer quotes from well-formed shell commands
The `StripQuotes` command parser uses this regex to detect "wrapped" commands:

```js
/^"(.+?)"$/.test(command)
```

This fires on any command string that both starts and ends with a `"` character — including perfectly valid shell commands like:

```
"/usr/local/bin/mytool" --flag "some value" --other "last arg"
```

The intent (supporting Windows where users write `"cmd with spaces"` as a single token) is reasonable, but the regex matches far too broadly. It cannot distinguish between:

- `"single wrapped token"` ← intended target
- `"/path/to/bin" --arg "quoted value"` ← well-formed shell command that coincidentally starts and ends with `"`

When the latter is matched, `slice(1, -1)` produces:

```
/path/to/bin" --arg "quoted value
```

…which is unparseable by the shell, causing a `not found` error (exit code 127).

**Suggested fix:** Only strip quotes when the command is a single quoted token (no inner quotes), e.g.:

```js
/^"([^"]+)"$/.test(command)   // no inner double quotes
```

or more precisely, only strip when the token contains no whitespace:

```js
/^"([^\s"]+)"$/.test(command)
```

This would safely handle `"mytool"` → `mytool` while leaving `"/bin/tool" --arg "val"` untouched.

**Version:** 9.2.1

jeffrey-takuma:
## Summary

Scope quote normalization to CLI input instead of the library API.

This moves the outer-quote normalization logic out of `lib/concurrently.ts` and into the CLI entrypoint, so well-formed shell commands keep their quotes when using the library API.

Closes #582

## Changes

- remove `StripQuotes` from `lib/concurrently.ts`
- add `bin/normalize-cli-command.ts` for CLI-only command normalization
- normalize CLI commands in `bin/index.ts` before calling `concurrently()`
- add coverage for CLI normalization in `bin/normalize-cli-command.spec.ts`
- add coverage for preserving quotes in the library API in `lib/concurrently.spec.ts`
- add a thin CLI integration test in `bin/index.spec.ts`

## Tests

- `pnpm test`


jeffrey-takuma:
Opened this PR for #582.

The test jobs are passing on the PR. The only failing check seems to be the coverage upload step, which is currently returning a Coveralls `504` timeout.

jeffrey-takuma:
@gustavohenke
I opened this PR to address #582, and it should be ready for review now. Could you take a look when you have a chance?

gustavohenke (approved):
Thank you, looks reasonable!
I've pushed some further, small changes

PR #604, opened 2026-07-28
fix: correctly output non-ASCII text on Windows

### LINKED ISSUE #302: Chinese garbled code on command line output
When there are Chinese characters on the command line, there will be garbled code.

The `scripts` in `package.json` file:

```json
  "scripts": {
    "test":"concurrently \"npm run print\"",
    "test2":"chcp 65001 && concurrently \"npm run print\"",
    "print":"echo 你好"
  }
```

Here is the output of `npm run test`:

```bash
PS D:\foo\example> npm run test

> example@0.0.0 test D:\foo\example
> concurrently "npm run print"

[0] 
[0] > example@0.0.0 print D:\foo\example
[0] > echo 你好
[0] 
[0] ���
[0] npm run print exited with code 0
```

To fix this, I currently add `chcp 65001` before `concurrently`. Here is the output of `npm run test2`:

```bash
PS D:\foo\example> npm run test2

> example@0.0.0 test2 D:\foo\example
> chcp 65001 && concurrently "npm run print"

Active code page: 65001
[0] 
[0] > example@0.0.0 print D:\foo\example
[0] > echo 你好
[0] 
[0] 你好
[0] npm run print exited with code 0
```

Environment:
- OS: Windows 11
- Node.js: 14.17.5
- concurrently: 7.0.0

DanielC000:
Closes #302.

I looked into this and your `chcp 65001 && <command>` idea from 2023 doesn't actually work if you inline it as a single command like that — `chcp` runs and reports success, but the CJK text is already garbled by the time it prints. cmd.exe seems to convert the whole `/c` command-line argument using the codepage that was active before `chcp` gets a chance to run. Confirmed on a real Windows 11 machine with raw byte output, not just what shows in a terminal.

What does work: setting the codepage on concurrently's own process once at startup — basically what the original reporter's own `chcp 65001 && concurrently ...` workaround was doing, since the codepage is a property of the console session and a freshly-piped child inherits whatever its parent console is set to. This does that in the CLI entry point only (not the library API, so nothing changes for anyone using concurrently programmatically), restores the original codepage on exit, and is a no-op if the codepage is already UTF-8 (increasingly common with Windows Terminal etc). Cost is one `chcp` probe at startup on Windows, and nothing at all on other platforms or under `--raw`.

Also tried cmd's own `/U` flag, since it doesn't need `chcp` at all. Doesn't help though — it only affects cmd.exe's own builtins, has zero effect on output from an external process like Node, and would have actively broken the currently-correct output from Node-based commands (most real usage), since it flips the encoding to UTF-16LE.

Before:

```
> npm run test
[0] ??
```

After:

```
> npm run test
[0] 你好
```

One thing I'm not sure about: since codepage is a per-console-session property, not per-process, this could in principle race with another codepage-sensitive tool running in the same terminal at the same time. It's scoped to the CLI only, and the restore hangs off the same completion path concurrently already uses for Ctrl+C, so I think the risk is small — but happy to put it behind a flag instead if you'd rather it not be automatic. WDYT?

### DIFF
--- a/bin/console-codepage.spec.ts
+++ b/bin/console-codepage.spec.ts
@@ -1,0 +1,83 @@
+import { describe, expect, it, vi } from 'vitest';
+
+import { ensureUtf8Codepage, SpawnSync } from './console-codepage.js';
+
+const win32 = { platform: 'win32' as const };
+const linux = { platform: 'linux' as const };
+
+const ok = (stdout: string) => ({ status: 0, stdout, error: undefined });
+const failed = () => ({ status: 1, stdout: '', error: undefined });
+
+describe('ensureUtf8Codepage()', () => {
+    it('does nothing on non-Windows platforms', () => {
+        const spawnSync = vi.fn();
+        ensureUtf8Codepage(spawnSync as unknown as SpawnSync, linux)();
+        expect(spawnSync).not.toHaveBeenCalled();
+    });
+
+    it('does nothing when the codepage is already UTF-8', () => {
+        const spawnSync = vi.fn().mockReturnValue(ok('Active code page: 65001.'));
+        const restore = ensureUtf8Codepage(spawnSync as unknown as SpawnSync, win32);
+
+        expect(spawnSync).toHaveBeenCalledTimes(1);
+
+        restore();
+        expect(spawnSync).toHaveBeenCalledTimes(1);
+    });
+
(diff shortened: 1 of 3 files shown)


gustavohenke (approved):
Excellent use of the [KISS principle](https://en.wikipedia.org/wiki/KISS_principle), lol

Thanks for fixing this!

PR #608, opened 2026-08-14
fix: expand wildcards from package.json5 when package.json is missing

### LINKED ISSUE #527: Using a `pnpm:` shortcut with a wildcard (*) does not work with JSON5 (package.json5)
pnmp supports using `package.json5` instead of the regular `package.json`, which is super useful to be able to add comments in the package file.

Concurrently allows specifying a [shortcut command](https://github.com/open-cli-tools/concurrently/blob/main/docs/cli/shortcuts.md) with a `*` wildcard like `concurrently "pnpm:lint:*"`, which would run all scripts in `package.json` that start with `lint:`.

This all works well when running with package.json, but when using packages.json5, it results in the following error:

```bash
$ pnpm test

> example-project@ test C:\.......\example-project
> concurrently --group pnpm:test:*

 ELIFECYCLE  Test failed. See above for more details.
```

This error occurs even if the file contents have not changed at all (i.e. still valid regular JSON) and only the file has been renamed to `.json5`

DSeaStar:
Fixes #527.

Wildcard expansion only read package.json. Fall back to package.json5 using the existing JSONC parser, same pattern as deno.jsonc.

### DIFF
--- a/docs/cli/shortcuts.md
+++ b/docs/cli/shortcuts.md
@@ -1,6 +1,6 @@
 # Command Shortcuts
 
-Package managers that execute scripts from a `package.json` or `deno.(json|jsonc)` file can be shortened when in concurrently.<br/>
+Package managers that execute scripts from a `package.json` / `package.json5` or `deno.(json|jsonc)` file can be shortened when in concurrently.<br/>
 The following are supported:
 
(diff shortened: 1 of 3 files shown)


gustavohenke (approved):
👍

PR #611, opened 2026-08-18
docs: update API example to ESM

### LINKED ISSUE #606: Update docs for new API usage (ESM style)

Bibin-VR:
concurrently v10 is pure ESM (type: module, and exports maps only ./dist/lib/index.js), so the README example no longer runs as written:

  const concurrently = require('concurrently');   // -> namespace object
  concurrently([...]);                            // TypeError: not a function

Verified against the published concurrently@10.0.5 on Node 22.

Switch the example to a named ESM import and replace __dirname, which does not exist in an ES module, with import.meta.dirname (Node >=20.11; package engines already require >=22).

Closes #606

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,9 @@ Check out documentation and other usage examples in the [`docs` directory](./doc
 Example:
 
 ```js
-const concurrently = require('concurrently');
+import path from 'node:path';
+import { concurrently } from 'concurrently';
+
 const { result } = concurrently(
   [
     'npm:watch-*',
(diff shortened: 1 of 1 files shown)


gustavohenke (approved):
Thanks

PR #616, opened 2026-09-03
fix: skip force-kill logging when no processes remain

imshunsuke:
## Summary

- avoid logging a force-kill event when no processes remain killable
- cover processes that exit after the initial termination signal

## Problem

When `killTimeout` is enabled, a process may exit normally after receiving the initial termination signal. The timeout callback currently treats the resulting empty command array as truthy and logs `Sending SIGKILL to 0 processes..`, even though no force kill occurs.

## Fix

Only log and send `SIGKILL` when the timeout callback finds at least one process that is still killable.

## Testing

- `pnpm vitest run --project unit lib/flow-control/kill-others.spec.ts`
- `CI=1 pnpm test`
- `pnpm exec vitest --coverage`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm format`
- `pnpm build`
- `pnpm test:smoke`

### DIFF
--- a/lib/flow-control/kill-others.spec.ts
+++ b/lib/flow-control/kill-others.spec.ts
@@ -134,6 +134,22 @@ it('force kills misbehaving processes after a timeout', () => {
     expect(commands[2].kill).toHaveBeenCalledTimes(1);
 });
 
+it('does not report a force kill when all processes exit before the timeout', () => {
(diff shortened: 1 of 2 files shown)


gustavohenke (approved):
lol, nice catch, LGTM

